PR #47, opened 2022-04-08
Feature/new-mutation-store

### LINKED ISSUE #45: Mutation share calculations assume `N` means reference causing problems
The nuc and aa mutation proportion API endpoints seem to assume that if a site is `N` it is reference and not mutated.

I don't think that assumption is good. It can cause problems when a consensus mutated site in a lineage is often `N` like for example 22813 in BA.1 (see issue https://github.com/lenaschimmel/sc2rf/issues/23).

It would be more natural to cancel `N`s out of the mutation share calculation.

So hypothetically speaking, at site 22813, if reference is A, and we have 40 `N`, 59 `T` and 1 `A` we should report mutation share of `T` as 59/60 = 98% rather than 59%. I think 98% is what people would assume the number means.

chaoran-chen:
This PR introduces a new in-memory store for the mutations which is also able to efficiently store positions at which the base/code is unknown/ambiguous.

The `aa-mutations` and `nuc-mutations` endpoints now calculate precise proportions (closes #45) and are much faster (roughly 50-200x faster). The small downside is that it needs more RAM (as it loads all mutation information into memory) and the program start-up takes a bit longer.


corneliusroemer:
Maybe there's another problem due to this PR? Website reloads all the time now. Thanks for implementing the feature.
https://github.com/cevo-public/cov-spectrum-website/issues/456

chaoran-chen:
Now, the new version is online! Hopefully, it' going to be stable.

PR #72, opened 2023-01-19
Fix warnings, add dependabot config

fengelniederhammer:


chaoran-chen (approved):
Looks good to me!

.github/dependabot.yml
```diff
@@ -0,0 +1,10 @@
+version: 2
```
fengelniederhammer, line 1: For this file to take effect, dependabot needs to be enabled for the repo. @chaoran-chen, is it enabled? https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuring-dependabot-version-updates#enabling-dependabot-version-updates

PR #85, opened 2023-01-23
#69: Usage and performance statistics

### LINKED ISSUE #69: Usage and performance statistics
It would be very useful to gather some usage and performance data. I suggest collecting the following data for each request:

- Date or date+hour+minute (or exact timestamp?)
- Endpoint
- Cache hit/miss
- Response time
- Variant

We get roughly 500,000 requests per day.

The easiest solution could be just writing the data to a log file? I'm open to ideas about the technical implementation and how we can easily get some statistics/plots!

JonasKellerer:
Resolves #69


server/build.gradle
```diff
@@ -43,6 +44,8 @@ dependencies {
     testImplementation 'org.junit.jupiter:junit-jupiter-engine'
     testImplementation 'org.junit.jupiter:junit-jupiter-params'
     testImplementation 'org.junit.jupiter:junit-jupiter-api'
+    testImplementation 'org.mockito:mockito-core:5.0.0'
+    testImplementation 'com.fasterxml.jackson.datatype:jackson-datatype-jsr310'
```
fengelniederhammer, line 48: - [ ] remove

server/src/main/java/ch/ethz/lapis/api/log/RequestContext.java
```diff
@@ -0,0 +1,20 @@
...
+
+@Component
+@RequestScope
+@Data
+public class RequestContext {
+    private LocalDateTime timestamp;
+    private Duration responseTimeInSeconds;
+    private String endpoint;
+    private boolean returnedDataFromCache = false;
+    private SampleFilter variantFilter;
```
chaoran-chen, line 19: this should better just be called `filter` as it includes more than the variant.

chaoran-chen (changes requested):
Thanks! This works already quite well. Just a few things:

The insertions and mutations (both AA and nuc) are logged in the decoded form, e.g. as `[{"gene":"S","position":501,"query":"YYY"}]` or `[{"gene":"S","position":501,"mutation":"Y","applyMaybe":false}]`. That's very unhandy and it would be better if they could be logged using the short and standardized notation. I.e., as `ins_S:501:YYY` and `S:501Y`.

It would also be great if the fields with a `null` could be omitted.

JonasKellerer:
The new log will look similar to this.

`{"timestamp":"2023-01-23T15:04:46.364506146","responseTimeInSeconds":0.007652409,"endpoint":"/v1/sample/aggregated","returnedDataFromCache":false,"filter":{"country":"Germany","nucMutations":[],"nucInsertions":[],"aaMutations":["S:501Y"],"aaInsertions":["ins_S:2345:EN"],"fields":["DATE","PANGOLINEAGE"]}}`

Is this, what you imagined?

chaoran-chen (changes requested):
The log looks great but something else went wrong.

If I open http://localhost:2345/v1/sample/aggregated?country=Switzerland&aaMutations=S:E484K,S:N501Y&fields=country, I get

```
{"errors":[],"info":{"apiVersion":1,"dataVersion":1665960786},"data":{"fields":["COUNTRY"],"samples":[{"country":"Switzerland","count":611}]}}
```

The wrapper around the samples is new. https://lapis.cov-spectrum.org/open/v1/sample/aggregated?country=Switzerland&aaMutations=S:E484K,S:N501Y&fields=country returns:

```
{"errors":[],"info":{"apiVersion":1,"dataVersion":1674178709,"deprecationDate":null,"deprecationInfo":null,"acknowledgement":null},"data":[{"country":"Switzerland","count":483}]}
```

---

Apart from that, a small request: it would be better to return the response time in milliseconds and as an integer.

chaoran-chen (changes requested):
One last request: It would be better to log the timestamp with the time zone or as the Unix timestamp so that it is actually clear which time it was without knowing the server config.

Otherwise, it looks good to me!

chaoran-chen (approved):
Great! Thank you!

PR #90, opened 2023-01-23
fix: CSV/TSV formatting is not working for mutations endpoints

### LINKED ISSUE #89: CSV/TSV formatting is not working for mutations endpoints
At the moment, https://lapis.cov-spectrum.org/open/v1/sample/aa-mutations?dataFormat=tsv returns an empty file. Returning JSON works and the endpoints `aggregated` and `details` also work.

The reason is that [SampleMutationsResponse.MutationEntry](https://github.com/GenSpectrum/LAPIS/blob/a471e80b96f071836b791b48add29746c7ee8b36/server/src/main/java/ch/ethz/lapis/api/entity/res/SampleMutationsResponse.java#L15) is a record but not a class with getter functions. Our self-written [CsvSerializer](https://github.com/GenSpectrum/LAPIS/blob/a471e80b96f071836b791b48add29746c7ee8b36/server/src/main/java/ch/ethz/lapis/api/entity/res/CsvSerializer.java#L58) however only looks for getters (methods starting with `get` or `is`) and extracts the field names from the method names.

chaoran-chen:
Closes #89

### DIFF
--- a/server/src/main/java/ch/ethz/lapis/api/entity/res/CsvSerializer.java
+++ b/server/src/main/java/ch/ethz/lapis/api/entity/res/CsvSerializer.java
@@ -45,26 +45,47 @@ public class CsvSerializer {
 
     public <T> String serialize(List<T> objects, Class<T> c, List<String> fields) {
         Map<String, Pair<String, Method>> allFields = new HashMap<>(); // The keys are lower-cased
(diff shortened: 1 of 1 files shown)


fengelniederhammer:
At least I don't know of any solution that immediately does the job. So far the code looks good to me (except maybe a missing test :) ).

PR #93, opened 2023-01-24
refactor: delete unused main method and related code

fengelniederhammer:


server/src/main/java/ch/ethz/lapis/core/ConfigurationManager.java
```diff
@@ -19,14 +20,11 @@ public class ConfigurationManager {
     /**
      * Loads the configuration from both the config file and from environment variables
      */
-    public <T extends Config> T loadConfiguration(Path configPath, Class<T> configClass, String programName)
+    public LapisConfig loadConfiguration(Path configPath, String programName)
         throws IOException {
         Map<String, ?> fileConfigMaps1 = readConfigMapsFromYaml(configPath);
         Map<String, ?> envConfigMaps1 = readConfigMapFromEnvironmentVariables();
```
fengelniederhammer, line 26: @chaoran-chen Is this actually still used? Or could we also clean up the mechanism to simply read the config from the file?

chaoran-chen (approved):
Looks good!

server/src/main/java/ch/ethz/lapis/core/ConfigurationManager.java
fengelniederhammer: Yes, we'll remove it in a separate PR then.

PR #94, opened 2023-01-25
refactor: simplify reading the config

fengelniederhammer:
Reading the config from environment variables is unused -> remove it.

Reading the config thus effectively becomes a one-liner, making snakeyaml obsolete, because we can simply use jackson with it's yaml module.

@chaoran-chen The config files for the productive servers need to be adapted, too, they are probably not stored in this repo, are they?


chaoran-chen (approved):
Looks good, thanks!

PR #103, opened 2023-02-01
Specify "fields" that should be returned by the /details endpont

### LINKED ISSUE #68: Specify "fields" that should be returned by the /details endpont
At the moment, the `details` endpoint returns all metadata information of the selected samples. Now, a `fields` parameter should be added that can be used to specify the fields that should be returned. It is similar to what we already have for the `aggregated` endpoint.

An example request would be: `https://lapis.cov-spectrum.org/open/v1/sample/details?limit=10&fields=date,country,genbankAccession`

JonasKellerer:
Resolves #68


chaoran-chen (approved):
It works as expected and the code looks good to me! Thanks!

PR #113, opened 2023-02-07
Fix: Charset wrongly interpreted (at least for TSV output)

### LINKED ISSUE #108: Charset wrongly interpreted (at least for TSV output)
When opening [this request](https://lapis.cov-spectrum.org/open/v1/sample/details?limit=10&division=Z%C3%BCrich&dataFormat=tsv) in Chrome, I see:

image

The response is UTF-8 (which is correct) but the browser does not seem to know it. I guess that we are missing something in the header. I tested this only for TSV outputs but imagine that the same problem also occurs for CSV and other plain text outputs.

(I'm putting this under prioritized because I assume that it's easy to fix but not because it's particularly important.)

JonasKellerer:
Resolves #108


chaoran-chen (approved):
Great!

PR #115, opened 2023-02-08
fix: update init.sql, insert some log statements

fengelniederhammer:


database/init.sql
```diff
@@ -381,4 +382,4 @@ on
   y_main_aa_sequence,
   y_main_aa_sequence_columnar,
   access_key
-to y_user;
+to lapis_proc;
```
chaoran-chen, line 385: This bloc can be removed because lapis_proc is already getting all the permission it needs [here](https://github.com/GenSpectrum/LAPIS/blob/7bae322f06ffb66788631d6c2b5c78e8fc21dae3/database/users.sql#L6)

chaoran-chen (commented):
Just one tiny change request. Otherwise, it can be merged :)

chaoran-chen (approved):

PR #119, opened 2023-02-14
include RKI sequences in open data instance

### LINKED ISSUE #114: ENH: Include `database=rki` sequences in open data
For about 10 days we've included RKI sequences in the ncov-ingest open data outputs.

However, these ~600k recent sequences are not showing in open cov-spectrum. I think that may be because these sequences do not have a genbank_accession - which covspectrum may use as primary key, because in the past every sequence had such an accession.

Now, the strain name is the only primary key. Alternatively, you could patch the metadata and fill the empty genbank_accession column with the RKI data's strain name, it's primary key suitable, something like IMS-123123-12312312.

fengelniederhammer:
resolves #114


server/src/test/java/ch/ethz/lapis/source/TestDataGenerator.java
```diff
@@ -0,0 +1,154 @@
...
+                }
+                i++;
+                outStream.write((line + "\n").getBytes());
+                rkiEntries.add(line.split("\t")[0]);
+                if (i > LIMIT) {
+                    break;
+                }
+            }
+            System.out.println("fertig mit Strain Namen: " + (System.currentTimeMillis() - start) + "ms");
+            System.out.println(rkiEntries.size());
```
fengelniederhammer, line 81: clean up

corneliusroemer:
Exciting! Let me know how/where I can test :)

chaoran-chen:
Nice, thanks for implementing it!

It's quite big and difficult to review it in detail (with all the field names, etc.). Should I maybe just run it on the server and see if everything works? With ba520ea4c92798f74ef1ea177afb252da9d93135 that I just pushed, the old data won't be deleted after importing and we can easily reverse.

fengelniederhammer:
> It's quite big and difficult to review it in detail

Did you review the individual commits? The second commit should be a lot easier to review alone.

chaoran-chen:
Ah, yes, that's much better. The code looks good! But there will be a problem with using the `nuc-sequence` and `aa-sequence` endpoints afterward because they use the GenBank Accession as key ([source](https://github.com/GenSpectrum/LAPIS/blob/5f61ef5fa17940bf8cbe7f320a2bfc0b563ac862/server/src/main/java/ch/ethz/lapis/api/SampleService.java#L348)).

We should switch the identifier for the fasta files to `strain`. As the endpoint is barely used (I checked the logs), it's fine to just do it.

fengelniederhammer:
Thanks for catching this! I also added the possibility to specify `fields=database` in the /aggregated and /details to stay consistent on that end.

- [ ] clean up commits before merge

server/src/main/java/ch/ethz/lapis/source/ng/NextstrainFileToDatabaseService.java
```diff
@@ -0,0 +1,664 @@
...
+                        statement.setString(5, currentHash);
+                        statement.addBatch();
+                        i++;
+                        if (i % 10000 == 0) {
+                            log.info("executing batch in updateSeqOriginalOrAligned for fasta entry i=" + i);
+                            Utils.executeClearCommitBatch(conn, statement);
+                        }
+                        if (i > 500000) {
+                            break;
+                        }
```
chaoran-chen, line 159: Don't forget to remove this ;)

fengelniederhammer: Removed it

chaoran-chen:
I ran the preprocessing pipeline on the server and it works well! One remaining issue is, however, that the RKI data do not contain a `host` but CoV-Spectrum only shows sequences from humans by default. I suggest that we set the `host` of RKI sequences to `Homo sapiens`.

@corneliusroemer, do you know whether it is safe to assume that the RKI sequences are from humans? Would it make sense to change the value for hosts on the Nextstrain side?

JonasKellerer:
> I ran the preprocessing pipeline on the server and it works well! One remaining issue is, however, that the RKI data do not contain a `host` but CoV-Spectrum only shows sequences from humans by default. I suggest that we set the `host` of RKI sequences to `Homo sapiens`.
> 
> @corneliusroemer, do you know whether it is safe to assume that the RKI sequences are from humans? Would it make sense to change the value for hosts on the Nextstrain side?

The host should now be set to `Homo sapiens` if it is null for the rki data. Could you @chaoran-chen check if the fix works?

server/src/main/java/ch/ethz/lapis/transform/TransformService.java
```diff
@@ -550,86 +551,111 @@ public void mergeFromPangolinAssignment() throws SQLException {
...
             }
         }
+
+        log.info("Finished final transforms");
+    }
+
+    private static void setHostForRkiData(Statement statement) throws SQLException {
+        String sql = """
+            update y_main_metadata_staging m
+            set host = "Homo sapiens"
```
chaoran-chen, line 572: Ah, wait, that's wrong! In SQL, the strings must be in single quotes.

chaoran-chen (changes requested):

chaoran-chen (approved):
Thanks! I tried the SQL query and it works. Green lights to merge from my side!

chaoran-chen:
Oh, wait, stop a moment..

chaoran-chen:
The commit "[tmp: avoid removing old data after importing for testing purposes](https://github.com/GenSpectrum/LAPIS/pull/119/commits/7b7f89dd8129a54940407baeca2407f8ca9455db)" is back. I actually deleted it two days ago. I'll delete it again and force-push.

chaoran-chen:
Okay. Done!

PR #158, opened 2023-02-28
docs: system scope and context chapter of arc42 documentation

### LINKED ISSUE #141: Documentation: Chapter System Scope And Context

JonasKellerer:
Resolves #141

Modeled after: https://docs.arc42.org/section-3/
Happy to hear your thoughts


docs/acr42/03_system_scope_and_context.md
```diff
@@ -0,0 +1,63 @@
...
+- one file for each gene
+
+metadata_and_quality_control.tsv
+- list of metadata and quality control
+
+insertions.tsv
+- list of nucleotide insertions and AA insertions for each sequence
+
+alias.tsv
+- list of pangolin aliases
```
JonasKellerer, line 54: Do we want to keep this here? The system needs to be reloaded, when the aliases change in this case. But is this more likely than that new data will arrive?

docs/arc42/12_glossary.md
```diff
@@ -4,5 +4,38 @@
...
+The following explains the notations for mutations.
+
+**Amino acid**
+
+The gene has to be provided for the AA mutation, since AAs only make sense within a gene.
+
+Example ORF_1a:G1234S translates to
+
+- in Gene: ORF_1a
+- AA mutation from G to S at position 1234
```
fengelniederhammer, line 26: Does the position actually start at 0 or at 1?

chaoran-chen: It starts at 1

fengelniederhammer: I think SILO currently starts at 0:

```
POST http://localhost:8081/query
Content-Type: application/json

{
  "action": {
    "type": "Mutations"
  },
  "filterExpression": {
    "type": "True"
  }
}
```
returns
```
{
  ...,
  "queryResult": [
    {
      "count": 45,
      "mutation": "A0-",
      "proportion": 0.9782608695652174
    },
    {
      "count": 44,
      "mutation": "T1-",
      "proportion": 0.9565217391304348
    },
...
]}
```
@Taepper we should check that.

Taepper: Good point. The filter should consider this (although the logic for that should probably be moved from the `getBitmap`-call to the parsing phase).
For the mutation distribution this should be definitely be changed. But the change should also happen when building the resulting string of {from_symbol}{position}{to_symbol}, not when returning the mutation struct.

PR #166, opened 2023-03-09
Integrate with silo

### LINKED ISSUE #122: Integrate with SILO
In the previously built `/aggregated` endpoint, send a hardcoded SILO query to SILO instead of returning a hardcoded response.

fengelniederhammer:
resolves #122 

This might look a bit overenginerred, but the neat thing about the client is: It can determine from the action type which response type it should expect.


lapis2/src/main/kotlin/org/genspectrum/lapis/silo/SiloClient.kt
```diff
@@ -0,0 +1,47 @@
...
+            .build()
+
+        val response =
+            try {
+                client.send(request, BodyHandlers.ofString())
+            } catch (exception: Exception) {
+                throw SiloException(exception.message, exception)
+            }
+
+        if (response.statusCode() >= 400) {
```
chaoran-chen, line 28: Why not check `!= 200`?

chaoran-chen (approved):
It looks good to me! Just have one small question below.

lapis2/src/main/kotlin/org/genspectrum/lapis/silo/SiloClient.kt
fengelniederhammer: Good question. I'll change it to `!= 200` ;)

PR #170, opened 2023-03-13
Docker for v2

### LINKED ISSUE #123: Build Docker container
Put LAPIS v2 into a docker container. Possibly also integrate with SILO containers via docker compose.

fengelniederhammer:
resolves #123 

Images can be found at https://github.com/GenSpectrum/LAPIS/pkgs/container/lapis-v2.
This uses the gradle task `bootBuildImage` that ships with Spring Boot.


chaoran-chen (approved):
I didn't know about `bootBuildImage` but it looks nice, probably better than our old, self-written, and naive images.

fengelniederhammer:
AFAIK it only exists since Spring Boot 3, i.e. it is not available in the old LAPIS. But there are other solutions to automate creating Docker images. People from TNG recommended [jib](https://github.com/GoogleContainerTools/jib). In any case the recommendation was to not maintain Dockerfiles yourself (I also didn't know that ;) ).

In principle one could also customize `bootBuildImage` a bit more, e.g. by explicitly specifying builder and runImage (https://docs.spring.io/spring-boot/docs/current/gradle-plugin/reference/htmlsingle/#build-image.customization). But as long as it works, we can leave the maintenance of that to Spring Boot and use their reasonable defaults.

PR #191, opened 2023-03-21
Implement SILO queries for simple mutation filters

### LINKED ISSUE #179: Implement SILO queries for simple mutation filters
Allow filtering for nucleotide mutations, e.g. `/aggregated?nucleotideMutations=G501C` and `/aggregated?nucleotideMutations=G501C,42T`.

JonasKellerer:
Resolves #179


lapis2/src/main/kotlin/org/genspectrum/lapis/model/AggregatedModel.kt
```diff
@@ -34,6 +35,25 @@ class AggregatedModel(private val siloClient: SiloClient) {
...
+        }
         return StringEquals(key, value)
     }
+
+    private fun generateNucleotideFilter(userInput: String): SiloFilterExpression {
+        val mutations = userInput.split(",")
+
+        return And(
+            mutations.map { mutationExpression ->
+                val match = Regex("(?<symbolFrom>^[A-Za-z]{0,})?(?<position>\\d{1,})(?<symbolTo>[A-Za-z]\$)")
```
chaoran-chen, line 49: A nucleotide mutation expression can also contain deletions represented by a `-`. See https://github.com/GenSpectrum/LAPIS/blob/e00d42f27ac4ae7215189c15c89b343534a860d9/server/src/main/antlr/ch/ethz/lapis/api/parser/VariantQuery.g4#L41

fengelniederhammer (changes requested):

fengelniederhammer (changes requested):
The Kotlin part looks good, but let's also add end to end tests :)

lapis2/src/test/kotlin/org/genspectrum/lapis/model/AggregatedModelTest.kt
```diff
@@ -99,6 +113,45 @@ class AggregatedModelTest {
...
+                        ),
+                    ),
+                ),
+            ),
+        )
+
+        @JvmStatic
+        fun getNucleotideMutationWithWrongSyntax() = listOf(
+            Arguments.of("A123"),
+            Arguments.of("123"),
```
chaoran-chen, line 150: These two are entirely valid! They look for sequences that have any mutation at position 123.

See https://cov-spectrum.org/about "How can I search a variant?"

lapis2/src/main/kotlin/org/genspectrum/lapis/model/AggregatedModel.kt
chaoran-chen: `.` is syntactic sugar for _not mutated_. Example: If the reference genome has an `A` at position 100, then `100.` is equivalent to `100A`.

I don't know whether SILO supports the syntax and I don't know whether it should. As it is purely for convenience, we might want to only support it in LAPIS. @Taepper, what do you think?

fengelniederhammer: SILO has the filters `NucleotideEquals` and `HasNucleotideMutation`. We implemented the mapping to `NucleotideEquals`. `100.` corresponds to `Not(HasNucleotideMutation)`, doesn't it?
But SILO also already supports `100.` by mapping it to `100<looked up from reference genome>`.

https://github.com/GenSpectrum/LAPIS-SILO/blob/main/src/silo/query_engine/query_engine.cpp#L135-L150

chaoran-chen: I think they are different when it comes to unknowns. By default, `HasNucleotideMutation` should only be true if a sequence has a confirmed mutation. If a sequence has an `N`, it will be evaluated as false. `Not(HasNucleotideMutation)` is then true. `.`, on the other hand, is true if a sequence has definitely the reference base.

If SILO already supports `.`, we should use it.

lapis2/src/test/kotlin/org/genspectrum/lapis/model/AggregatedModelTest.kt
JonasKellerer: Good catch. I added tests, that show that both work. As far as I can tell, Silo currently does not support "any mutation" besides using the symbol `N` (which I used as a workaround), but this not the intended behavior, is it?

siloLapisTests/test/aggregatedQueries/regionIsEurope.json
```diff
@@ -0,0 +1,9 @@
+{
+  "testCaseName": "country equals Switzerland",
```
chaoran-chen, line 2: The name is wrong? Shouldn't it be "region equals Europe"?

chaoran-chen (changes requested):
Found a wrong test case name. Otherwise, it looks good to me! Thanks! (Feel free to merge after fixing it)

siloLapisTests/test/aggregatedQueries/regionIsEurope.json
fengelniederhammer: Thanks for catching this, I fixed it.

PR #202, opened 2023-03-30
fix: correctly map PangoLineageEquals filters

### LINKED ISSUE #201: Fix PangoLineage filter
A PangoLineage filter needs the field `include_sublineages`.

fengelniederhammer:
resolves #201 

This depends on https://github.com/GenSpectrum/LAPIS-SILO/pull/64.

### DIFF
--- a/lapis2/src/main/kotlin/org/genspectrum/lapis/model/AggregatedModel.kt
+++ b/lapis2/src/main/kotlin/org/genspectrum/lapis/model/AggregatedModel.kt
@@ -25,19 +25,28 @@ class AggregatedModel(private val siloClient: SiloClient) {
             SiloQuery(
                 SiloAction.aggregated(),
                 And(
-                    filterParameter.map { convertToSiloFilter(it.key, it.value) },
+                    filterParameter.map { mapToSiloFilter(it.key, it.value) },
                 ),
             ),
         )
     }
 
-    fun convertToSiloFilter(key: String, value: String) = when (key) {
-        "pangoLineage" -> PangoLineageEquals(value)
-        "nucleotideMutations" -> generateNucleotideFilter(value)
+    fun mapToSiloFilter(key: String, value: String) = when (key) {
+        "pangoLineage" -> mapToPangoLineageFilter(value)
+        "nucleotideMutations" -> mapToNucleotideFilter(value)
         else -> StringEquals(key, value)
     }
 
-    private fun generateNucleotideFilter(userInput: String): SiloFilterExpression {
+    private fun mapToPangoLineageFilter(value: String) = when {
+        value.endsWith(".*") -> PangoLineageEquals(value.substringBeforeLast(".*"), includeSublineages = true)
+        value.endsWith('*') -> PangoLineageEquals(value.substringBeforeLast('*'), includeSublineages = true)
+        value.endsWith('.') -> throw IllegalArgumentException(
+            "Invalid pango lineage: $value must not end with a dot. Did you mean '$value*'?",
+        )
+        else -> PangoLineageEquals(value, includeSublineages = false)
+    }
+
+    private fun mapToNucleotideFilter(userInput: String): SiloFilterExpression {
         val mutations = userInput.split(",")
 
         return And(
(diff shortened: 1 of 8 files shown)


lapis2/src/main/kotlin/org/genspectrum/lapis/model/AggregatedModel.kt
```diff
@@ -25,19 +25,24 @@ class AggregatedModel(private val siloClient: SiloClient) {
...
-        "nucleotideMutations" -> generateNucleotideFilter(value)
+    fun mapToSiloFilter(key: String, value: String) = when (key) {
+        "pangoLineage" -> mapToPangoLineageFilter(value)
+        "nucleotideMutations" -> mapToNucleotideFilter(value)
         else -> StringEquals(key, value)
     }
 
-    private fun generateNucleotideFilter(userInput: String): SiloFilterExpression {
+    private fun mapToPangoLineageFilter(value: String) = when {
+        value.endsWith('*') -> PangoLineageEquals(value.substringBeforeLast('*'), includeSublineages = true)
```
chaoran-chen, line 41: There is one subtlety here. It is possible to write both `BA.1*` and `BA.1.*`, they are considered completely equivalent.  Both should be mapped to "`BA.1` including sublineages".

chaoran-chen (changes requested):

lapis2/src/main/kotlin/org/genspectrum/lapis/model/AggregatedModel.kt
fengelniederhammer: Should we handle this in LAPIS or in SILO?

My gut feeling says that we should push this to SILO and ignore it in LAPIS. SILO will need to do some validations and checks on this anyway.

chaoran-chen: Do you mean that SILO should also accept `BA.1.` as a query? This could also be an option.

Then, we should at least add a test case in LAPIS to ensure (and document) that `BA.1.*` is a valid value.

fengelniederhammer: What does `BA.1.` actually mean? Is it the same as `BA.1`?

chaoran-chen: At the moment, `BA.1.` is not accepted (nothing will be found). However, `BA.1.*` is valid and means exactly the same as `BA.1*`. As SILO does not use the `*` syntax but requires an object with `{pango lineage, include sub-lineage}`, we have to decide how to map.

Suggestion 1:

- `BA.1` -> `{pango lineage: BA.1, include sub-lineage: false}`
- `BA.1.` -> Throw error message?
- `BA.1*` -> `{pango lineage: BA.1, include sub-lineage: true}`
- `BA.1.*` -> `{pango lineage: BA.1, include sub-lineage: true}`

Suggestion 2:

- `BA.1` -> `{pango lineage: BA.1, include sub-lineage: false}`
- `BA.1.` -> ???
- `BA.1*` -> `{pango lineage: BA.1, include sub-lineage: true}`
- `BA.1.*` -> `{pango lineage: BA.1., include sub-lineage: true}`

The constraint is, to maintain backward compatibility, that `BA.1*` and `BA.1.*` both work for LAPIS users and that they return the same.

I think that suggestion 1 is the easiest/most straight-forward but suggestion 2 is also okay.

chaoran-chen: To be precise: `BA.1.` currently leads to somewhat inconsistent behavior. Used in a "simple" query, it does not cause an error but also does not return any sample. Used in an advanced variant query, it causes an error.

fengelniederhammer: Suggestion 1 sounds reasonable, but I don't like that LAPIS needs to know quite a lot about the structure of a field. This might be hard to generalize.

chaoran-chen (approved):
Looks good, thanks!

JonasKellerer (approved):

PR #222, opened 2023-04-10
SQL for chat

chaoran-chen:
I added a small SQL(-like) interface. Its sole purpose is to evaluate SQL queries that my small ChatGPT-based application generates. Only very basic SQL queries are accepted. The queries are **not** evaluated in conformity with the SQL standards but in a way that should produce a result that is (hopefully) close enough and useful for the chat application.

In the "virtual" model, there are three database tables: `metadata`, `aa_mutations`, and `nuc_mutations`. In `metadata`, there is one row per sequence. In the mutations tables, there is one row per sequence-mutation combination. All the tables have one column per metadata field plus one column per position in the nucleotide sequence and AA sequence. To filter for `2134T`, we would query `where nuc_2134 = 'T'` and for `S:862P`, the query is `where aa_S_862 = 'P'`. The mutations tables have, in addition, the column `mutation` containing the name of the mutation. Some example queries can be found [here](https://github.com/GenSpectrum/cov-spectrum-server/blob/729f925b817f73f1e5099a73320943e9007c5afc/src/main/kotlin/ch/ethz/covspectrum/chat/OpenAIClient.kt#L41).

This is a very naive implementation. After finishing the first version of LAPIS v2, we should discuss whether we want to add proper support for (a subset of) SQL. (cc @Taepper, FYI)


swantan:
Hi, I would like to find out the query prefix for SARS-CoV-2 sequences from GISAID. Currently, I am aware of the prefix https://lapis.cov-spectrum.org/open/v1 if we were to query an endpoint from the open data/GenBank. Thank you and I would greatly appreciate your help!

chaoran-chen:
Hi @swantan, very unfortunately, we are not allowed to share our GISAID instance due to GISAID's terms and conditions.

PR #238, opened 2023-04-25
Access key authentication

### LINKED ISSUE #218: Access key authentication
For non-open instances such as the GISAID one, we only allow access if an access key is provided. 

- It should be possible to specify the [data openness level](https://github.com/GenSpectrum/LAPIS/blob/d7d50c9bac37ab4128f2db73aa2d2d060f6e0246/server/src/main/java/ch/ethz/lapis/api/controller/v1/SampleController.java#L78) in the config. There should be two openness levels: `open` and `gisaid`.
- If the openness level is `gisaid`, the server should only allow access with a valid access key. In the current LAPIS, the access keys are stored in the PostgreSQL database. As we probably won't have an external relational database in LAPIS v2, we can also just keep them in a file.

fengelniederhammer:
resolves #218

### DIFF
--- a/lapis2/src/main/kotlin/org/genspectrum/lapis/config/DatabaseConfig.kt
+++ b/lapis2/src/main/kotlin/org/genspectrum/lapis/config/DatabaseConfig.kt
@@ -4,6 +4,7 @@ data class DatabaseConfig(val schema: DatabaseSchema)
 
 data class DatabaseSchema(
     val instanceName: String,
+    val opennessLevel: OpennessLevel,
     val metadata: List<DatabaseMetadata>,
     val primaryKey: String,
     val features: List<DatabaseFeature> = emptyList(),
@@ -12,3 +13,8 @@ data class DatabaseSchema(
 data class DatabaseMetadata(val name: String, val type: String)
 
 data class DatabaseFeature(val name: String)
+
+enum class OpennessLevel {
+    OPEN,
+    GISAID,
+}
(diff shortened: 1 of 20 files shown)


JonasKellerer (commented):
Nice. I like it :) 
Can we test the behavior also in e2e tests?

lapis2/src/main/kotlin/org/genspectrum/lapis/config/DatabaseConfig.kt
```diff
@@ -4,11 +4,17 @@ data class DatabaseConfig(val schema: DatabaseSchema)
...
 data class DatabaseSchema(
     val instanceName: String,
+    val opennessLevel: OpennessLevel,
     val metadata: List<DatabaseMetadata>,
     val primaryKey: String,
     val features: List<DatabaseFeature> = emptyList(),
 )
 
-data class DatabaseMetadata(val name: String, val type: String)
+data class DatabaseMetadata(val name: String, val type: String, val unique: Boolean = false)
```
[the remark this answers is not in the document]
fengelniederhammer: `valuesAreUnique` might be even better

lapis2/src/main/kotlin/org/genspectrum/lapis/config/DatabaseConfig.kt
```diff
@@ -4,11 +4,17 @@ data class DatabaseConfig(val schema: DatabaseSchema)
...
 
 data class DatabaseFeature(val name: String)
+
+enum class OpennessLevel {
+    OPEN,
+    GISAID,
```
chaoran-chen, line 19: I think that it's better if we name it `PROTECTED` (and should we later have different protected modes, we should try to find general names).

Also, would this be a good place to document briefly what the two levels mean?

fengelniederhammer: done

PR #280, opened 2023-07-03
feature: handle int and float columns

fengelniederhammer:


lapis2/src/main/kotlin/org/genspectrum/lapis/model/SiloFilterExpressionMapper.kt
```diff
@@ -52,6 +49,10 @@ class SiloFilterExpressionMapper(
                 Filter.DateBetween -> mapToDateBetweenFilter(siloColumnName, values)
                 Filter.NucleotideSymbolEquals -> mapToNucleotideFilter(values[0].value)
                 Filter.VariantQuery -> mapToVariantQueryFilter(values[0].value)
+                Filter.IntEquals -> mapToIntEqualsFilter(siloColumnName, values)
+                Filter.IntBetween -> mapToIntBetweenFilter(siloColumnName, values)
+                Filter.FloatEquals -> mapToFloatEqualsFilter(siloColumnName, values)
+                Filter.FloatBetween -> mapToFloatBetweenFilter(siloColumnName, values)
```
JonasKellerer, line 55: Only for consistency, one time we use `values` and in the corresponding function `values[0].value`, on the other time we use `values[0].value` as an input.

JonasKellerer (approved):

lapis2/src/main/kotlin/org/genspectrum/lapis/model/SiloFilterExpressionMapper.kt
fengelniederhammer: Those that get `values[0].value` passed only expect one value. The others have to deal with possibly multiple values.

PR #285, opened 2023-07-05
feat: add fields to aggregated endpoint

### LINKED ISSUE #216: add `fields` field to `/aggregated`
Make it possible to specify by which fields SILO should aggregate in the aggregated action. SILO already supports this, but we have to make LAPIS pass the information to SILO.

JonasKellerer:
Resolves #216


lapis2/src/main/kotlin/org/genspectrum/lapis/OpenApiDocs.kt
```diff
@@ -33,6 +35,29 @@ fun buildOpenApiSchema(sequenceFilterFields: SequenceFilterFields, databaseConfi
...
+                        ),
+                    ),
+            ).addSchemas(
+                RESPONSE_SCHEMA_AGGREGATED,
+                Schema<String>()
+                    .type("object")
+                    .description("aggregated sequence data")
+                    .required(listOf("count"))
+                    .properties(
+                        properties +
```
fengelniederhammer, line 56: This also contains `accessKey`, but it should not.

fengelniederhammer, line 56: Should we add a description here to every field that it is present if and only if it was specified in `fields` in the request?

fengelniederhammer (approved):

fengelniederhammer (changes requested):
Let's fix the access key thing, otherwise it looks good.

fengelniederhammer (approved):

PR #296, opened 2023-07-12
feat: implement amino acid mutation filters

### LINKED ISSUE #287: Filter by Amino Acid Mutations
After https://github.com/GenSpectrum/LAPIS-SILO/pull/159 is merged, we can extend LAPIS to also filter by amino acid mutations:
Valid examples:
* `/aggregated?aaMutations=S:501Y`
* `/aggregated?aaMutations=s:501Y``
* `/aggregated?aaMutations=S:N501Y`

See https://github.com/GenSpectrum/LAPIS/blob/d6747f9360f404fc9be7136e1bab7ed62dd6591d/server/src/main/java/ch/ethz/lapis/api/entity/AAMutation.java#L43 for the current behavior.

fengelniederhammer:
resolves #287
This also fixed nucleotide mutations: They need to be passed as a string array instead of a comma separated list


lapis2/src/test/kotlin/org/genspectrum/lapis/model/VariantQueryFacadeTest.kt
```diff
@@ -23,7 +26,7 @@ class VariantQueryFacadeTest {
 
     @Test
     fun `given a complex variant query then map should return the corresponding SiloQuery`() {
-        val variantQuery = "300G & (400- | 500B) & !600 & MAYBE(700B | 800-) & [3-of: 123A, 234T, 345G] & A.1.2.3*"
+        val variantQuery = "300G & (400- | 500B) & !600. & MAYBE(700B | 800-) & [3-of: 123A, 234T, 345G] & A.1.2.3*"
```
JonasKellerer, line 29: 600 should be also a valid variant query. Why was this changed here?

lapis2/src/test/kotlin/org/genspectrum/lapis/model/VariantQueryFacadeTest.kt
```diff
@@ -85,12 +88,21 @@ class VariantQueryFacadeTest {
...
+
+        val result = underTest.map(variantQuery)
+
+        assertThat(result, equalTo(HasNucleotideMutation(null, 400)))
     }
 
     @Test
     fun `given a variantQuery with an 'And' expression then map should return the corresponding SiloQuery`() {
-        val variantQuery = "300G & 400"
+        val variantQuery = "300G & 400-"
```
JonasKellerer, line 105: Same here with 400 :)

JonasKellerer (approved):
Nice :)

lapis2/src/test/kotlin/org/genspectrum/lapis/model/VariantQueryFacadeTest.kt
fengelniederhammer: Yes, but it doesn't mean `NucleotideSymbolEquals(null, 600, "-")` anymore. It's translated to a `HasNucleotideMutation`. I changed it such that I don't have to modify the output by much. But actually it doesn't matter. We can also keep the original version here and expect a `HasNucleotideMutation`

siloLapisTests/test/aggregatedQueries/noMutationAt23270.json
```diff
@@ -1,11 +1,11 @@
 {
   "testCaseName": "no mutation at 23063",
   "lapisRequest": {
-    "nucleotideMutations": "23063."
+    "nucleotideMutations": ["230."]
```
chaoran-chen, line 4: The test case name and the file name are not in sync with the query anymore.

I'd also like to suggest a second test case for "no mutation at 230" which is to query for `nucleotideMutations: [230C]`, it should return the same number.

chaoran-chen (commented):
One small comment, otherwise, everything looks good to me!

siloLapisTests/test/aggregatedQueries/noMutationAt23270.json
fengelniederhammer: Do you mean an explicit query for "is equal to the symbol in the reference genome"? I think I'd leave that for SILO, as that logic is handled there.

PR #317, opened 2023-08-04
185 separate preprocessing and starting the api

JonasKellerer:
Update LAPIS so it can work with https://github.com/GenSpectrum/LAPIS-SILO/pull/190


fengelniederhammer:
superseded by https://github.com/GenSpectrum/LAPIS/pull/327

PR #325, opened 2023-08-09
feat: wrap response data to match LAPIS 1 structure (#324)

### LINKED ISSUE #324: Adapt response structure of LAPIS 2 to LAPIS 1
LAPIS 2 currently simply returns the data as they are.

LAPIS 1 returns some additional information and the data wrapped in a "data" field:
```
{
    "errors": [],
    "info": {
        "apiVersion": 1,
        "dataVersion": 1691410888,
        "deprecationDate": null,
        "deprecationInfo": null,
        "acknowledgement": null
    },
    "data": [
        {
            "date": "2023-07-16",
            "count": 1
        },
        ...
    ]
}
```

Wrap the response data in a data field, so that we can easily extend it later with further information without breaking changes.

fengelniederhammer:
resolves #324

### DIFF
--- a/lapis2/src/main/kotlin/org/genspectrum/lapis/OpenApiDocs.kt
+++ b/lapis2/src/main/kotlin/org/genspectrum/lapis/OpenApiDocs.kt
@@ -79,24 +79,28 @@ fun buildOpenApiSchema(sequenceFilterFields: SequenceFilterFields, databaseConfi
                 )
                 .addSchemas(
                     AGGREGATED_RESPONSE_SCHEMA,
-                    Schema<String>()
-                        .type("object")
-                        .description(
-                            "Aggregated sequence data. " +
-                                "If fields are specified, then these fields are also keys in the result. " +
-                                "The key 'count' is always present.",
-                        )
-                        .required(listOf(COUNT_PROPERTY))
-                        .properties(getAggregatedResponseProperties(metadataFieldSchemas(databaseConfig))),
+                    lapisResponseSchema(
+                        Schema<String>()
+                            .type("object")
+                            .description(
+                                "Aggregated sequence data. " +
+                                    "If fields are specified, then these fields are also keys in the result. " +
+                                    "The key 'count' is always present.",
+                            )
+                            .required(listOf(COUNT_PROPERTY))
+                            .properties(getAggregatedResponseProperties(metadataFieldSchemas(databaseConfig))),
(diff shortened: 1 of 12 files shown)


lapis2/src/main/kotlin/org/genspectrum/lapis/OpenApiDocs.kt
```diff
@@ -106,6 +110,13 @@ fun buildOpenApiSchema(sequenceFilterFields: SequenceFilterFields, databaseConfi
         )
 }
 
+private fun lapisResponseSchema(dataSchema: Schema<Any>?) =
```
TobiasKampmann, line 113: optional?

lapis2/src/test/kotlin/org/genspectrum/lapis/auth/ProtectedDataAuthorizationTest.kt
```diff
@@ -53,13 +56,15 @@ class ProtectedDataAuthorizationTest(@Autowired val mockMvc: MockMvc) {
     fun `given no access key in POST request to protected instance, then access is denied`() {
         mockMvc.perform(postRequestWithBody(""))
             .andExpect(MockMvcResultMatchers.status().isForbidden)
```
TobiasKampmann, line 58: Just an idea: I like the import of the content() matcher. Can we do the same for status()?

lapis2/src/test/kotlin/org/genspectrum/lapis/auth/ProtectedDataAuthorizationTest.kt
```diff
@@ -53,13 +56,15 @@ class ProtectedDataAuthorizationTest(@Autowired val mockMvc: MockMvc) {
...
-            .andExpect(MockMvcResultMatchers.content().contentType(MediaType.APPLICATION_JSON))
+            .andExpect(content().contentType(MediaType.APPLICATION_JSON))
             .andExpect(
-                MockMvcResultMatchers.content().json(
+                content().json(
                     """
```
TobiasKampmann, line 62: The error string appears ~6 times here and apparently is always the same, i.e., we dont test other errors here. We could extract the error string. Not only we would shorten the tests by a bit, the error that is tested here can be on the top of the file....

TobiasKampmann (approved):
Looks great!

PR #341, opened 2023-08-22
feat: make url prefix configurable

fengelniederhammer:
This is necessary when deploying on a server with a proxy in between that remaps paths (e.g. example.server/lapis/aggregated -> 127.0.0.1/aggregated)


lapis2/docker-compose.yml
```diff
@@ -4,7 +4,7 @@ services:
     image: ghcr.io/genspectrum/lapis-v2:${LAPIS_TAG}
     ports:
       - "8080:8080"
-    command: --silo.url=http://silo:8081 --lapis.databaseConfig.path=databaseConfig.yaml
+    command: --silo.url=http://silo:8081 --lapis.database-config.path=databaseConfig.yaml
```
JonasKellerer, line 7: What is the reason behind this naming change?

JonasKellerer (approved):

lapis2/docker-compose.yml
JonasKellerer: Seems to be kebab-case for application properties.

TobiasKampmann (approved):
As far as I understand it, seems fine

fengelniederhammer:
Closing in favor of #343

PR #342, opened 2023-08-22
enable /aggregated to return the data as CSV

### LINKED ISSUE #217: enable `/aggregated` to return the data as CSV
As a user, I want to be able to specify the `dataFormat` CSV.

Allowed values are CSV, TSV and JSON. The default (if not specified) is JSON.

JonasKellerer:
Resolves #217

### DIFF
--- a/lapis2/src/main/kotlin/org/genspectrum/lapis/controller/LapisController.kt
+++ b/lapis2/src/main/kotlin/org/genspectrum/lapis/controller/LapisController.kt
@@ -37,6 +37,7 @@ const val AMINO_ACID_MUTATIONS_SCHEMA = "AminoAcidMutations"
 const val ORDER_BY_FIELDS_SCHEMA = "OrderByFields"
 const val LIMIT_SCHEMA = "Limit"
 const val OFFSET_SCHEMA = "Offset"
+const val FORMAT_SCHEMA = "DataFormat"
 
 const val DETAILS_ENDPOINT_DESCRIPTION = "Returns the specified metadata fields of sequences matching the filter."
 const val AGGREGATED_ENDPONT_DESCRIPTION = "Returns the number of sequences matching the specified sequence filters"
@@ -95,6 +96,12 @@ class LapisController(
         )
         @RequestParam
         offset: Int? = null,
+        @Parameter(
+            schema = Schema(ref = "#/components/schemas/$FORMAT_SCHEMA"),
+            description = "The format of the response.",
+        )
+        @RequestParam
+        dataFormat: String? = null,
     ): LapisResponse<List<AggregationData>> {
         val request = SequenceFiltersRequestWithFields(
             sequenceFilters?.filter { !SPECIAL_REQUEST_PROPERTIES.contains(it.key) } ?: emptyMap(),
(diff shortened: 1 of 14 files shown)


fengelniederhammer (changes requested):
The GET endpoints misses the dataFormat in the swagger UI (I guess I also missed that for the details endpoint).

lapis2/src/main/kotlin/org/genspectrum/lapis/controller/LapisController.kt
```diff
@@ -90,6 +96,12 @@ class LapisController(
         )
         @RequestParam
         offset: Int? = null,
+        @Parameter(
+            schema = Schema(ref = "#/components/schemas/$FORMAT_SCHEMA"),
+            description = "The format of the response.",
```
fengelniederhammer, line 101: Let's maybe reuse the description that we already have in the schema in `OpenApiDocs`? Maybe we could add there that the parameter takes precedence over the header.

lapis2/src/main/kotlin/org/genspectrum/lapis/response/SiloResponse.kt
```diff
@@ -8,11 +8,38 @@ import com.fasterxml.jackson.databind.JsonNode
...
+    override fun asHeader() = fields.keys.plus(COUNT_PROPERTY).toTypedArray()
+}
+
+data class DetailsData(val map: Map<String, JsonNode>) : Map<String, JsonNode> by map, CsvRecord {
+    override fun asArray() = values.map { it.asText() }.toTypedArray()
+    override fun asHeader() = keys.toTypedArray()
+}
+
+@JsonComponent
+class DetailsDataSerializer : JsonSerializer<DetailsData>() {
```
fengelniederhammer, line 27: Do we really need this? I think it might work without it.

lapis2/src/main/kotlin/org/genspectrum/lapis/response/SiloResponse.kt
```diff
@@ -8,11 +8,38 @@ import com.fasterxml.jackson.databind.JsonNode
...
+        gen.writeEndObject()
+    }
+}
+
+@JsonComponent
+class DetailsDataDeserializer : JsonDeserializer<DetailsData>() {
+    override fun deserialize(p: JsonParser, ctxt: DeserializationContext): DetailsData {
+        val node = p.readValueAsTree<JsonNode>()
+        val fields = node.fields().asSequence().associate { it.key to it.value }
+        return DetailsData(fields)
```
fengelniederhammer, line 40: I would have expected something in the spirit of (since it worked without an explicit deserializer before, too):
```suggestion
        return DetailsData(p.readValueAsTree<Map<String, JsonNode>>())
```

lapis2/src/main/kotlin/org/genspectrum/lapis/controller/LapisController.kt
```diff
@@ -381,20 +525,24 @@ class LapisController(
...
     ): String {
-        return getDetailsAsCsv(request, Delimiter.TAB)
+        return getResponseAsCsv(request, Delimiter.TAB, siloQueryModel::getDetails)
     }
 
-    private fun getDetailsAsCsv(request: SequenceFiltersRequestWithFields, delimiter: Delimiter): String {
+    private fun getResponseAsCsv(
+        request: SequenceFiltersRequestWithFields,
+        delimiter: Delimiter,
+        getResponse: (request: SequenceFiltersRequestWithFields) -> List<CsvRecord>,
```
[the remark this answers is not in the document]
JonasKellerer: A complete generic is not possible due to `requestContext.filter = request`. However `<Request : CommonSequenceFilters> getResponseAsCsv` works.

lapis2/src/main/kotlin/org/genspectrum/lapis/response/SiloResponse.kt
JonasKellerer: Good idea. The readValueAsTree can not get the Map directly. One need to use 
`return DetailsData(p.readValueAs(object : TypeReference<Map<String, JsonNode>>() {}))`

Issue #375, opened 2023-09-13
Deploy documentation


chaoran-chen:
Deploy the new Astro-based documentation on the test server, for the instance created in #371

PR #387, opened 2023-09-20
build(lapis2-docs): build Docker image

fengelniederhammer:
targets #375

### DIFF
--- a/lapis2-docs/Dockerfile
+++ b/lapis2-docs/Dockerfile
@@ -1,0 +1,15 @@
+FROM node:lts-alpine as build-deps
+
+WORKDIR /app
+
+RUN apk update && apk add curl
+
+COPY . .
+ENV CONFIG_FILE=/config/database_config.yaml
+
+EXPOSE 3000
+VOLUME /config
+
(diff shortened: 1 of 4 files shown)


lapis2-docs/Dockerfile
```diff
@@ -0,0 +1,15 @@
...
+
+COPY . .
+ENV CONFIG_FILE=/config/database_config.yaml
+
+EXPOSE 3000
+VOLUME /config
+
+HEALTHCHECK --start-period=60s CMD curl --fail --silent localhost:3000 | grep "Welcome to LAPIS" || exit 1
+
+CMD npm ci && npm run build && npm run preview -- --host 0.0.0.0 --port 3000
```
fengelniederhammer, line 15: executing "npm ci" only here reduces the image size (180 MB compared to ~700 MB)

JonasKellerer (approved):
Nice. Thanks. 

I think we need some kind of readme to describe how to build it.

fengelniederhammer:
Writing a README sounds like a good idea.

PR #414, opened 2023-10-04
chore: group dependabot updates

fengelniederhammer:


TobiasKampmann (approved):
As this worked in pathoplexus...

PR #424, opened 2023-10-11
Simplify config in docker

fengelniederhammer:
This avoids double configuration:
* Mounting files into the container
* Telling LAPIS where one mounted the files

Instead, just mount the files to the correct default location


JonasKellerer (approved):
Nice. Thank you.

PR #482, opened 2023-12-09
Add info to lapis response

### LINKED ISSUE #480: Let LAPIS return an info object alongside the data
Currently each endpoint returns an object, similar to `{"data": "someData"}`, where someData can be the results of an aggregated, mutation or details request. 

The returned object should be extended, such that it also contains an Info object, similar to

`{"data": "someData, "info": {"dataVersion": "1234"}`

The info object should contain for now only the dataVersion, which is also returned in the response header.
The info object should only be added if the request does not fail.

JonasKellerer:
Resolves #480

### DIFF
--- a/lapis2/src/main/kotlin/org/genspectrum/lapis/openApi/OpenApiDocs.kt
+++ b/lapis2/src/main/kotlin/org/genspectrum/lapis/openApi/OpenApiDocs.kt
@@ -27,6 +27,7 @@ import org.genspectrum.lapis.controller.OFFSET_PROPERTY
 import org.genspectrum.lapis.controller.ORDER_BY_PROPERTY
 import org.genspectrum.lapis.request.AminoAcidInsertion
 import org.genspectrum.lapis.request.AminoAcidMutation
+import org.genspectrum.lapis.request.LAPIS_DATA_VERSION_HEADER
 import org.genspectrum.lapis.request.NucleotideInsertion
 import org.genspectrum.lapis.request.NucleotideMutation
 import org.genspectrum.lapis.request.OrderByField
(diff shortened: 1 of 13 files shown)


lapis2/src/main/kotlin/org/genspectrum/lapis/request/LapisInfo.kt
```diff
@@ -1,19 +1,29 @@
...
 import org.springframework.web.servlet.mvc.method.annotation.ResponseBodyAdvice
 
-@Component
-@RequestScope
-class DataVersion {
-    var dataVersion: String? = null
+data class LapisInfo(var dataVersion: String? = null)
+
+@JsonComponent
+class LapisInfoSerializer : JsonSerializer<LapisInfo>() {
```
fengelniederhammer, line 21: We probably don't need this, do we?

lapis2/src/main/kotlin/org/genspectrum/lapis/request/LapisInfo.kt
```diff
@@ -29,6 +39,15 @@ class ResponseBodyAdviceDataVersion(private val dataVersion: DataVersion) : Resp
...
     ): Any? {
         response.headers.add(LAPIS_DATA_VERSION_HEADER, dataVersion.dataVersion)
+
+        if (body is LapisResponse<*>) {
+            return LapisResponse(body.data, LapisInfo(dataVersion.dataVersion))
+        }
+
+        if (body is LapisErrorResponse) {
+            return LapisErrorResponse(body.error, LapisInfo(dataVersion.dataVersion))
+        }
```
fengelniederhammer, line 49:
```suggestion
        when (body) {
            is LapisResponse<*> -> body.info.dataVersion = dataVersion.dataVersion
            is LapisErrorResponse -> body.info.dataVersion = dataVersion.dataVersion
        }
```
or else make `LapisInfo.dataVersion` a `val`

fengelniederhammer (changes requested):
`info` should also be added to the response in the OpenAPI specs.

lapis2/src/main/kotlin/org/genspectrum/lapis/openApi/OpenApiDocs.kt
```diff
@@ -249,8 +249,19 @@ private fun lapisResponseSchema(dataSchema: Schema<Any>) =
...
         ),
-    ).required(listOf("data"))
+    ).required(listOf("data", "info"))
+
+private fun infoResponseSchema() =
+    Schema<Any>().type("object")
+        .description("Information about LAPIS.")
+        .properties(
+            mapOf(
+                "dataVersion" to Schema<String>().type("string").description("The data version of the data in SILO.")
```
fengelniederhammer, line 261: `LapisResponseAnnotation` already has a description for that. I think we should unify that. And maybe point out, that it's actually the same value?

fengelniederhammer (approved):

Issue #376, opened 2023-09-13
Adapt documentation


chaoran-chen:
Check the documentation, fix errors, and add content/create issues for changes.

PR #496, opened 2023-12-12
chore(docs): we don't need internationalization #376

fengelniederhammer:
@chaoran-chen was there a specific reason why you added this collection? Currently Astro shows the warning:
"11:08:36 [WARN] [content] The i18n collection is defined but no content/i18n folder exists in the content directory. Create a new folder for the collection, or check your content configuration file for typos."


chaoran-chen:
I think that it was part of the Starlight template and I didn't add it intentionally.

JonasKellerer (approved):

PR #497, opened 2023-12-13
fix: consider only servlet URL when checking auth

fengelniederhammer:
Don't consider the full "requestURI" as this might contain a prefix that a proxy sets.


JonasKellerer (approved):
LGTM

PR #505, opened 2023-12-15
feat: allow lower case for mutations

JonasKellerer:


fengelniederhammer (commented):
* Should insertions be case insensitive, too?
* For consistency, we should probably implement the same thing in the variant queries, too.

JonasKellerer:
There are now tickets for the insertions, and the variant queries: #508 #509

lapis2/src/main/kotlin/org/genspectrum/lapis/request/AminoAcidMutation.kt
```diff
@@ -26,15 +26,15 @@ data class AminoAcidMutation(val gene: String, val position: Int, val symbol: St
             return AminoAcidMutation(
                 gene,
                 position,
-                matchGroups["symbolTo"]?.value,
+                matchGroups["symbolTo"]?.value?.toString()?.uppercase(),
```
fengelniederhammer, line 29: IntelliJ claims that the `toString` is redundant
```suggestion
                matchGroups["symbolTo"]?.value?.uppercase(),
```

fengelniederhammer: Same for NucleotideMutation

fengelniederhammer (approved):

PR #520, opened 2023-12-19
fix: set default min proportion

fengelniederhammer:
See https://github.com/GenSpectrum/LAPIS-SILO/pull/234


JonasKellerer (approved):
LGTM

PR #558, opened 2024-01-09
552 docs mutation filters

### LINKED ISSUE #552: Docs: Mutation filters
Fix the "Mutation filters" page (some symbols are not shown correctly). Maybe improve it. Make sure it has the same content as https://github.com/GenSpectrum/LAPIS/wiki/4.4-Mutation-filters and then delete that page.

fengelniederhammer:
resolves #552


lapis2-docs/src/components/MutationFilters/NucleotideMutations.astro
```diff
@@ -0,0 +1,16 @@
+---
+import { getReferenceGenomes, isMultiSegmented } from '../../reference_genomes';
+import SegmentNames from './SegmentNames.astro';
+
+const referenceGenomes = getReferenceGenomes();
+const multiSegmented = isMultiSegmented(referenceGenomes);
+---
+
+<p>
+    A nucleotide mutation has the format <code>{multiSegmented && '<segment>:'}{'<position><base>'}</code>.
```
JonasKellerer, line 10: Could we not also have `<base><position><base>`, where the first base is optional, or do we leave it out, since we do not take it into account?
(Same for amino acids)

JonasKellerer (approved):

lapis2-docs/src/components/MutationFilters/NucleotideMutations.astro
fengelniederhammer: We simply ignore the "from base", so I think I would simply leave it out here...

PR #559, opened 2024-01-09
docs: update mutation description

### LINKED ISSUE #554: Update mutations description
A query of mutations (nucleotide or amino acid) returns the proportion of samples having the mutation. However, ambiguous reads are not counted to the total number of sequences. Thus, a request for mutations can have
`...
{ count: 10, proportion: 0.01, mutation: asdf }, 
{ count 20, proportion: 0.01, mutation: uiae },
...
`
This is confusing, since one would expect count divided by proportion gives the total number of sequences. This total number should be the same for all mutations. But this is not the case, due to the ambiguous reads. So, this should be reflected in the endpoint description, and the schema of the swagger.

JonasKellerer:
Resolves #554

### DIFF
--- a/lapis2/src/main/kotlin/org/genspectrum/lapis/controller/ControllerDescriptions.kt
+++ b/lapis2/src/main/kotlin/org/genspectrum/lapis/controller/ControllerDescriptions.kt
@@ -1,44 +1,46 @@
 package org.genspectrum.lapis.controller
 
-const val DETAILS_ENDPOINT_DESCRIPTION = "Returns the specified metadata fields of sequences matching the filter."
-const val AGGREGATED_ENDPOINT_DESCRIPTION = "Returns the number of sequences matching the specified sequence filters"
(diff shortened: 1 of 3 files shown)


lapis2/src/main/kotlin/org/genspectrum/lapis/controller/ControllerDescriptions.kt
```diff
@@ -4,10 +4,14 @@ const val DETAILS_ENDPOINT_DESCRIPTION = "Returns the specified metadata fields
 const val AGGREGATED_ENDPOINT_DESCRIPTION = "Returns the number of sequences matching the specified sequence filters"
 const val NUCLEOTIDE_MUTATION_ENDPOINT_DESCRIPTION =
     "Returns the number of sequences matching the specified sequence filters, " +
-        "grouped by nucleotide mutations."
+        "grouped by nucleotide mutations. Additionally, the relative frequency of each mutation is returned. It is" +
```
fengelniederhammer, line 7: Using a multiline string might make this more readable.

fengelniederhammer (approved):
LGTM

PR #560, opened 2024-01-09
feat(docs): transfer ambiguous symbols explanation #551

### LINKED ISSUE #551: Docs: Ambiguous symbols
Migrate https://github.com/GenSpectrum/LAPIS/wiki/4.5-Ambiguous-symbols to the docs.

fengelniederhammer:
resolves #551 

transferring https://github.com/GenSpectrum/LAPIS/wiki/4.5-Ambiguous-symbols

### DIFF
--- a/lapis2/src/test/kotlin/org/genspectrum/lapis/request/AminoAcidMutationTest.kt
+++ b/lapis2/src/test/kotlin/org/genspectrum/lapis/request/AminoAcidMutationTest.kt
@@ -80,6 +80,14 @@ class AminoAcidMutationTest {
                     "\"gENe1:123A\"",
                     AminoAcidMutation("gene1", 123, "A"),
                 ),
+                Arguments.of(
+                    "\"MAYBE(gene1:123A)\"",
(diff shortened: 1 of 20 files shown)


lapis2/src/test/kotlin/org/genspectrum/lapis/request/AminoAcidMutationTest.kt
```diff
@@ -80,6 +80,10 @@ class AminoAcidMutationTest {
                     "\"gENe1:123A\"",
                     AminoAcidMutation("gene1", 123, "A"),
                 ),
+                Arguments.of(
+                    "\"MAYBE(gene1:123A)\"",
```
JonasKellerer, line 84: Only out of curiosity: when I type `MAYBE(gene1:G123A)` this would also be a valid query? If yes, I would add it here to the test. (I mean we ignore the symbol before 123, but it would be nice to have it explicit here)

JonasKellerer (approved):
LGTM. Only one small question.

PR #567, opened 2024-01-11
Expose configuration

### LINKED ISSUE #555: Expose configuration
* In the docs, create a page that simply shows the DB config and the reference genome of the underlying LAPIS instance.
* In LAPIS, create two endpoints
  * one that returns the DB config
  * one that returns the reference genome

JonasKellerer:
Resolves #555

### DIFF
--- a/lapis2-docs/src/components/ReferenceGenome.astro
+++ b/lapis2-docs/src/components/ReferenceGenome.astro
@@ -1,0 +1,9 @@
+---
+import { getReferenceGenomes } from '../reference_genomes';
+
+const referenceGenome = getReferenceGenomes();
+---
+
+<div>
+    <pre class='overflow-x-auto'><code>{JSON.stringify(referenceGenome, null, 2)}</code></pre>
+</div>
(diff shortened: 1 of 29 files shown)


lapis2-docs/src/components/ReferenceGenome.astro
```diff
@@ -6,5 +6,5 @@ const referenceGenome = getReferenceGenomes();
 ---
 
 <div>
-    <pre class='overflow-x-auto'><code>{objectToPrettyJson(referenceGenome)}</code></pre>
+    <pre class='overflow-x-auto'><code>{JSON.stringify(referenceGenome, null, 2)}</code></pre>
```
fengelniederhammer, line 9: reminder: squash before merge

lapis2-docs/src/components/ReferenceGenome.astro
```diff
@@ -0,0 +1,10 @@
+---
+import { getReferenceGenomes } from '../reference_genomes';
+import { objectToPrettyJson } from '../utils/objectToPrettyJson';
```
fengelniederhammer, line 3: We don't need that import anymore

fengelniederhammer (approved):
LGTM

PR #574, opened 2024-01-15
fix authentication, don't document authentication for now

### LINKED ISSUE #553: Docs: Authentification
Add a section on how the "PROTECTED" mode of LAPIS works.

fengelniederhammer:
Also requests like "fields=[<primaryKey>]" should be considered to return detailed data. We decided not to document authentication for now. The current implementation is special logic required for the GISAID instance. It might be removed in the future. Anyway, if someone else need authentication on LAPIS, we should implement something better.

resolves #553

## PR Checklist
- [ ] ~~The implemented feature is covered by an appropriate end-to-end test.~~


JonasKellerer (approved):
LGTM

PR #576, opened 2024-01-15
fix: variant query with nextcladePangolineageQuery

JonasKellerer:
resolves #

## PR Checklist
- [ ] ~~All necessary documentation has been adapted.~~
- [ ] The implemented feature is covered by an appropriate end-to-end test.


JonasKellerer:
Currently, we have no e2e tests checking this. Do we need them?

JonasKellerer:
I added, that our Listener now fails, when there is more than one element in the stack.

I think we could rewrite the template, to say test.

fengelniederhammer:
Looks good.

PR #587, opened 2024-01-22
feat: implement a request id to trace requests

### LINKED ISSUE #586: Implement request IDs to trace errors
* allow consumers to pass a request ID in the request
* if none is provided, generate one
* include it in every log message of the request
* include it in error messages.

fengelniederhammer:
resolves #586

## PR Checklist

### DIFF
--- a/lapis2-docs/src/content/docs/concepts/request-id.mdx
+++ b/lapis2-docs/src/content/docs/concepts/request-id.mdx
@@ -1,0 +1,23 @@
+---
+title: Request Id
+description: Request Id
+---
+
+LAPIS uses a request id to identify requests.
+This is useful for debugging and logging purposes.
+
+:::note
+If you report an error that occurred in LAPIS, please include the request id.
+This will greatly simplify identifying the problem in our log files.
+:::
+
+You can provide a request id in the request as an HTTP header `X-Request-Id`.
+This is useful if you want to correlate requests in your own log files with the LAPIS log files
+to track problems across systems.
+
+:::caution
+If you use the `X-Request-Id` header, make sure that the value is unique.
+:::
+
+If you don't specify a request id, LAPIS will generate one for you.
+It is contained in the response header `X-Request-Id` and in `info` in the response body.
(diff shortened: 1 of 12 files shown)


lapis2/src/main/kotlin/org/genspectrum/lapis/logging/RequestId.kt
```diff
@@ -0,0 +1,43 @@
...
+import java.util.UUID
+
+@Component
+@RequestScope
+class RequestIdContext {
+    var requestId: String? = null
+}
+
+private const val REQUEST_ID_MDC_KEY = "RequestId"
+private const val HIGH_PRECEDENCE_BUT_LOW_ENOUGH_TO_HAVE_REQUEST_SCOPE_AVAILABLE = -100
```
JonasKellerer, line 21: Is there docu we could link?

lapis2-docs/src/content/docs/concepts/request-id.mdx
```diff
@@ -0,0 +1,23 @@
...
+If you report an error that occurred in LAPIS, please include the request id.
+This will greatly simplify identifying the problem in our log files.
+:::
+
+You can provide a request id in the request as an HTTP header `X-Request-Id`.
+This is useful if you want to correlate requests in your own log files with the LAPIS log files
+to track problems across systems.
+
+:::caution
+If you use the `X-Request-Id` header, make sure that the value is unique.
```
JonasKellerer, line 19: What happens if it is not unique?

JonasKellerer (approved):
LGTM.
I only struggle with the scope. The request Id is for us very useful, but not for the enduser (which has no access to the LAPIS logs). Do we need to inform the user, that they can provide a request id? It seems they would not use that feature?

lapis2-docs/src/content/docs/concepts/request-id.mdx
fengelniederhammer: Well, we loose a bit of traceability...

lapis2/src/main/kotlin/org/genspectrum/lapis/logging/RequestId.kt
fengelniederhammer: Not really. This was try and error. `-500` doesn't work.

lapis2-docs/src/content/docs/concepts/request-id.mdx
JonasKellerer: Okay. Thats fine for me.

PR #593, opened 2024-01-23
feat: amino acid and nucleotide insertion filters are case-insensitive

### LINKED ISSUE #592: Mutation queries should be case-insensitive
Mutation and insertion queries should be case-insensitive. At the moment, it seems that `ins_S:11:A` works but `INS_S:11:A` doesn't. Both should work.

JonasKellerer:
resolves #592 

## PR Checklist


fengelniederhammer (approved):
LGTM

Issue #604, opened 2024-01-29
Docs: Document special request properties


fengelniederhammer:
such as `limit`, `downloadAsFile` and `offset`

PR #605, opened 2024-01-29
feat: implement downloadAsFile

### LINKED ISSUE #599: `downloadAsFile` parameter
The same as #26, just for v2.

The filename of the downloaded file should be sensible (compare https://github.com/GenSpectrum/LAPIS/commit/95ba6d051475ce3784cf6cc06a05c925ec39ce6c#diff-c17d35b23549a3aa75a4c0cc6cd5b464c65b56769a7d85c6bac15ae9f3da66dbR110)

fengelniederhammer:
resolves #599

## PR Checklist
- ~~[ ] All necessary documentation has been adapted.~~ #604

### DIFF
--- a/lapis2/src/main/kotlin/org/genspectrum/lapis/openApi/DownloadAsFileParameterCustomizer.kt
+++ b/lapis2/src/main/kotlin/org/genspectrum/lapis/openApi/DownloadAsFileParameterCustomizer.kt
@@ -1,0 +1,22 @@
+package org.genspectrum.lapis.openApi
+
+import io.swagger.v3.oas.models.OpenAPI
+import io.swagger.v3.oas.models.parameters.Parameter
+import org.genspectrum.lapis.controller.DOWNLOAD_AS_FILE_PROPERTY
+import org.springdoc.core.customizers.OpenApiCustomizer
+import org.springframework.stereotype.Component
+
+@Component
+class DownloadAsFileParameterCustomizer : OpenApiCustomizer {
+    override fun customise(openApi: OpenAPI) {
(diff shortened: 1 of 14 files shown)


lapis2/src/main/kotlin/org/genspectrum/lapis/openApi/DownloadAsFileParameterCustomizer.kt
```diff
@@ -0,0 +1,22 @@
+package org.genspectrum.lapis.openApi
```
JonasKellerer, line 1: Sadly I still get the problem with swagger, so I could not test it. There is an issue for it #610

siloLapisTests/test/aggregated.spec.ts
```diff
@@ -196,4 +195,12 @@ null	Switzerland	2
     expect(result.status).equals(200);
     expect(result.headers.get('lapis-data-version')).to.match(/\d{10}/);
   });
+
+  it('should return the data with Content-Disposition when asking for download', async () => {
```
JonasKellerer, line 199: Should we add this test to a common spec file, since we need it for all other routes? (There are others also like this, dataVersion etc.)

JonasKellerer (approved):
LGTM :)

siloLapisTests/test/aggregated.spec.ts
fengelniederhammer: I'll do it in a separate, follow-up PR :+1:

fengelniederhammer: -> #613

PR #620, opened 2024-02-01
fix(docs): schema for uploaded config

fengelniederhammer:
## PR Checklist
- ~~[ ] The implemented feature is covered by an appropriate end-to-end test.~~

### DIFF
--- a/lapis2-docs/src/components/ConfigGenerator/MetadataWizard.tsx
+++ b/lapis2-docs/src/components/ConfigGenerator/MetadataWizard.tsx
@@ -105,7 +105,7 @@ export function MetadataEditModal({ index, metadata }: { index: number; metadata
 
     const [metadataType, setMetadataType] = useState<MetadataType>(metadata.type);
     const [metadataName, setMetadataName] = useState(metadata.name);
-    const [generateIndex, setGenerateIndex] = useState(metadata.generateIndex);
+    const [generateIndex, setGenerateIndex] = useState(!!metadata.generateIndex);
 
     const handleUpdateMetadata = () => {
         updateMetadata({ ...metadata, name: metadataName, type: metadataType, generateIndex }, index);
--- a/lapis2-docs/src/components/ConfigGenerator/configContext.tsx
+++ b/lapis2-docs/src/components/ConfigGenerator/configContext.tsx
@@ -24,49 +24,37 @@ export type MetadataType = z.infer<typeof metadataTypeSchema>;
 export const metadataSchema = z.object({
     name: z.string(),
     type: metadataTypeSchema,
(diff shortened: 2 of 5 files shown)


lapis2-docs/src/components/ConfigGenerator/configContext.tsx
```diff
@@ -24,49 +24,37 @@ export type MetadataType = z.infer<typeof metadataTypeSchema>;
...
-    instanceName: string;
-    opennessLevel: OpennessLevel;
-    metadata: Metadata[];
-    primaryKey: string;
-    dateToSortBy: string;
-    partitionBy: string;
-    tableColumns: string[];
-    features: Feature[];
-};
+export type Schema = z.infer<typeof schemaSchema>;
```
JonasKellerer, line 51: ;) schemaSchema

lapis2-docs/src/components/ConfigGenerator/MetadataWizard.tsx
```diff
@@ -105,7 +105,7 @@ export function MetadataEditModal({ index, metadata }: { index: number; metadata
 
     const [metadataType, setMetadataType] = useState<MetadataType>(metadata.type);
     const [metadataName, setMetadataName] = useState(metadata.name);
-    const [generateIndex, setGenerateIndex] = useState(metadata.generateIndex);
+    const [generateIndex, setGenerateIndex] = useState(!!metadata.generateIndex);
```
JonasKellerer, line 108: Shouldn't we also change the type of `generateIndex`? Or do we want to have explicitly in the config `generateIndex: false`. Then we would need to adapt this here and change the `handleSetMetadataType` accordingly

JonasKellerer (approved):
LGTM

lapis2-docs/src/components/ConfigGenerator/configContext.tsx
fengelniederhammer: Well, the schema in this file is: call the zod schema `somethingSchema` and the type `Something`. If you have a better suggestion than `schemaSchema`, I'm happy to hear it :D

JonasKellerer: I have no better idea, but funny nonetheless. ;)

lapis2-docs/src/components/ConfigGenerator/MetadataWizard.tsx
fengelniederhammer: I checked in SILO and the code looks like it doesn't care whether `generateIndex` is present or not. So I would go for this simple implementation here (which always adds `generateIndex`)

PR #641, opened 2024-02-13
fix: Sending unknown value in fields returns incomprehensible error

### LINKED ISSUE #619: Sending unknown value in `fields` returns incomprehensible error
`http://localhost:8080/sample/aggregated?accessKey=testAggregatedDataAccessKey&fields=DOES_NOT_EXIST` returns

```json
{
    "type": "about:blank",
    "title": "Bad Request",
    "status": 400,
    "detail": "Failed to convert 'fields' with value: 'DOES_NOT_EXIST'",
    "instance": "/sample/aggregated"
}
```

Adapt the mechanism to return something that can be understood by the user. Actually there is a proper error message, but Spring overwrites it.

JonasKellerer:
resolves #619 

## PR Checklist
- [ ] The implemented feature is covered by an appropriate end-to-end test.


siloLapisTests/test/aggregated.spec.ts
```diff
@@ -104,6 +104,18 @@ describe('The /aggregated endpoint', () => {
...
+  it('should return bad request for non existing fields on GET request', async () => {
+    const urlParams = new URLSearchParams({
+      fields: 'notAField',
+    });
+
+    const result = await getAggregated(urlParams);
+
+    expect(result.status).equals(400);
+    const resultJson = await result.json();
+    expect(resultJson.error.detail).to.include('Unknown field: notAField, known values are [primaryKey,');
```
fengelniederhammer, line 116: Actually I think we should also be able to see this in a unit test now? Should we add one?

fengelniederhammer (approved):
Nice :+1:

PR #643, opened 2024-02-13
download zstd and gzip compressed data

### LINKED ISSUE #600: Download zstd compressed data
For all endpoints, it should be possible to download the data as a Zstandard-compressed archive.

fengelniederhammer:
resolves #600

## PR Checklist

This PR includes:
* compression via Zstd and Gzip
  * via the `Accept-Encoding` header
  * or via the request property `compression`
* adapting the filename when `downloadAsFile=true` such that the file ending reflects the compression format
* The `Content-Encoding` header is set accordingly


lapis2/build.gradle
fengelniederhammer, line 1: - [ ] reminder: Squash before merge

lapis2/src/test/kotlin/org/genspectrum/lapis/controller/MockData.kt
```diff
@@ -0,0 +1,308 @@
...
+            Switzerland	42	3.14
+            Switzerland	43	null
+        """.trimIndent(),
+    )
+
+    private val nucleotideMutations = MockDataCollection.create(
+        siloQueryModelMockCall = { it::computeNucleotideMutationProportions },
+        modelData = listOf(
+            NucleotideMutationResponse(
+                "sequenceName:1234",
```
JonasKellerer, line 205: Do we want to add the symbolFrom and symbolTo, since this is used in many tests? Or is this the more difficult version to handle?

lapis2/src/main/kotlin/org/genspectrum/lapis/controller/CompressionFilter.kt
```diff
@@ -0,0 +1,141 @@
...
+    override fun getOutputStream() = servletOutputStream
+}
+
+class CompressingServletOutputStream(
+    private val outputStream: ServletOutputStream,
+    compression: Compression,
+) : ServletOutputStream() {
+    private val compressingStream = compression.compressionOutputStreamFactory(outputStream)
+
+    override fun write(b: Int) {
```
JonasKellerer, line 124: Is something special, or can we name it "byte" here?

lapis2/src/main/kotlin/org/genspectrum/lapis/openApi/Schemas.kt
```diff
@@ -79,7 +81,12 @@ If none if provided in the request, LAPIS will generate one.
 
 const val DOWNLOAD_AS_FILE_DESCRIPTION = """
 Set to true to make your browser trigger a download instead of showing the response content by setting the
-'Content-Disposition' header to 'attachment'.
+'$CONTENT_DISPOSITION' header to 'attachment'.
+"""
+
+const val COMPRESSION_DESCRIPTION = """
+Optionally set this to return the response compressed in the specified format.
+Alternatively, you can set the '$ACCEPT_ENCODING' header to the respective value.
```
JonasKellerer, line 89: When I try the compression of gzip in the browser I get the same json back. Is this intended? If so, it is strange to observe no change, when using the swagger. Maybe we could add something to the description?

JonasKellerer (approved):
LGTM

lapis2/src/test/kotlin/org/genspectrum/lapis/controller/MockData.kt
fengelniederhammer: Actually it doesn't matter, because we don't parse the response, but I'll add it for clarity.

lapis2/src/main/kotlin/org/genspectrum/lapis/controller/CompressionFilter.kt
fengelniederhammer: Ah no, I forgot to rename it. IntelliJ took this from the function that it overwrites.

lapis2/src/main/kotlin/org/genspectrum/lapis/openApi/Schemas.kt
chaoran-chen: What happens if you additionally set `downloadAsFile=true`? Is it showing or saving an uncompressed file (i.e., it downloads and then directly uncompresses)?

fengelniederhammer: The browser automatically handles gzip compressed data. From what I observed, most requests that a browser issues have `Accept-Encoding: gzip, deflate, br` set, and the response is decompressed without a user noticing it. Is that what you mean?

chaoran-chen: Is it possible to allow the user to get an actual .gz file downloaded to their disk using the browser?

PR #648, opened 2024-02-15
test: visit each page on its own in test for relative links

### LINKED ISSUE #603: Docs: Test each doc page separately for relative link tests
Currently, the e2e test loops through all pages and clicks all relative links in one test. This needs a lot of time. Therefore we increased the timeout of playwright. 

Change this, so we have a test for each page. Then reduce the timeout.

JonasKellerer:
resolves #603 

## PR Checklist
~~- [ ] The implemented feature is covered by an appropriate end-to-end test.~~

### DIFF
--- a/lapis2-docs/tests/docs.spec.ts
+++ b/lapis2-docs/tests/docs.spec.ts
@@ -1,49 +1,101 @@
 import { expect, Page, test } from '@playwright/test';
 import { baseUrl } from './queryGenerator.page';
 
-const gettingStartedPages = ['Introduction', 'Generate your request'];
+type DocumentationPage = {
+    title: string;
+    relativeUrl: string;
+};
 
-const referencesPages = [
-    'Introduction',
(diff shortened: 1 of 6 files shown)


lapis2-docs/tests/docs.spec.ts
```diff
@@ -55,13 +107,16 @@ const pages = [
 ];
 
 test.describe('The documentation', () => {
-    test('should show all expected pages via next buttons and all relative links should work', async ({ page }) => {
-        await page.goto(baseUrl);
+    pages.forEach((documentationPage) => {
```
fengelniederhammer, line 110:
```suggestion
    pages.forEach((documentationPage, indexOfCurrentPage) => {
```
and pass that to `findNeighbouringPages`, which saves us the weird computation there.

fengelniederhammer (approved):
Thanks, looks good :)

PR #649, opened 2024-02-15
fix: don't set Transfer-Encoding twice

fengelniederhammer:
Some proxies such as Traefik will throw an error when the Transfer-Encoding header is present twice. We shouldn't set the header ourselves, but leave that to Tomcat (which also makes sure that the chunking is done correctly). Instead, prevent setting a content length when the response is compressed.

related: #600 

resolves #

## PR Checklist


JonasKellerer (approved):
LGTM

PR #650, opened 2024-02-16
feat: return bad request error when unknown compression format in request

### LINKED ISSUE #647: Throw error for unknown compression format
If the request contains an unknown value for the `compression` property, LAPIS should return an error.

(Note: there should be no error when an unknown value is specified in the corresponding header).

fengelniederhammer:
resolves #647 

## PR Checklist


JonasKellerer (approved):
LGTM

PR #657, opened 2024-02-20
fix: return valid gzip when LAPIS returns an error response

### LINKED ISSUE #656: bad request - compressed
Apparently, when SILO returns a 400 and LAPIS was requested to return compressed data, LAPIS sends out compressed data, but not the corresponding header. Investigate and eventually fix.

fengelniederhammer:
Apparently the following happened:
* The gzip stream eagerly writes the gzip header to the underlying stream (in its constructor)
* an error occurs
* Spring's error handling resets the response buffer (erasing the gzip header)
* The exception handler writes a messages, which gets gzip compressed.

Fix by: Delay initializing the gzip output stream until something is written to it.

resolves #656

## PR Checklist


JonasKellerer (approved):
LGTM

PR #664, opened 2024-02-22
refactor: flatten mapped variant queries

fengelniederhammer:
people sometimes send very long variant queries that resulted in very deeply nested expressions

## PR Checklist
- ~~[ ] The implemented feature is covered by an appropriate end-to-end test.~~

### DIFF
--- a/lapis2/src/test/kotlin/org/genspectrum/lapis/model/SiloFilterExpressionMapperTest.kt
+++ b/lapis2/src/test/kotlin/org/genspectrum/lapis/model/SiloFilterExpressionMapperTest.kt
@@ -583,8 +583,8 @@ class SiloFilterExpressionMapperTest {
                     ),
                     And(
                         And(
-                            NucleotideSymbolEquals(null, 300, "G"),
                             NucleotideSymbolEquals(null, 400, "A"),
+                            NucleotideSymbolEquals(null, 300, "G"),
                         ),
                     ),
                 ),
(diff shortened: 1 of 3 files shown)


lapis2/src/test/kotlin/org/genspectrum/lapis/model/SiloFilterExpressionMapperTest.kt
```diff
@@ -583,8 +583,8 @@ class SiloFilterExpressionMapperTest {
                     ),
                     And(
                         And(
-                            NucleotideSymbolEquals(null, 300, "G"),
                             NucleotideSymbolEquals(null, 400, "A"),
+                            NucleotideSymbolEquals(null, 300, "G"),
```
JonasKellerer, line 587: Do we care that the order which is sent to SILO is reversed from the input?

JonasKellerer (approved):
LGTM.

Nice idea to flatten it here. (But I thought that SILO will do something similar).

lapis2/src/test/kotlin/org/genspectrum/lapis/model/SiloFilterExpressionMapperTest.kt
fengelniederhammer: The order should not matter.

fengelniederhammer:
SILO also flattens it, so this is mostly a minor optimization and making it easier to read for humans that check the logs.

PR #679, opened 2024-02-28
feat: set Content-Type header to application/gzip|zstd when the compression property in the request was set

### LINKED ISSUE #665: Fix compression headers
Currently, when we go to `aggregated?compression=gzip`, we get a response with

```
Content-Encoding: gzip
Content-Type: application/json
```

Similarly, if choosing `?compression=zstd`, one get a response with a `Content-Encoding: zstd`. Opening in a browser, this causes an error:

image

I think that the headers are actually quite wrong. If someone requests a compressed file, then the content is not JSON (or TSV or FASTA) but it's a GZIP or ZSTD file. I.e., there shouldn't be a `Content-Encoding` and the `Content-Type` should be `application/gzip` or `
application/zstd` (I checked that both are [official IANA media types](https://www.iana.org/assignments/media-types/media-types.xhtml)) or maybe just `application/octet-stream`.

fengelniederhammer:
resolves  #665

## PR Checklist

### DIFF
--- a/lapis2/src/main/kotlin/org/genspectrum/lapis/CorsConfiguration.kt
+++ b/lapis2/src/main/kotlin/org/genspectrum/lapis/CorsConfiguration.kt
@@ -1,7 +1,7 @@
 package org.genspectrum.lapis
 
-import org.genspectrum.lapis.openApi.REQUEST_ID_HEADER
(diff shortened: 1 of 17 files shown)


JonasKellerer (approved):
LGTM and works in my browsers.

Issue #658, opened 2024-02-21
Support order by random


fengelniederhammer:
Depends on https://github.com/GenSpectrum/LAPIS-SILO/issues/293

fengelniederhammer:
https://github.com/GenSpectrum/LAPIS/pull/681#issuecomment-1973268543

Make sure that "order by random" queries are never cached.

PR #681, opened 2024-03-01
feat: implement a request cache for mutation and insertion queries

### LINKED ISSUE #137: Implement a request cache
Cache requests to reduce load on SILO.

Details:
- the cache should be invoked for:
    - /nucleotideMutations
    - /aminoAcidMutations
    - /nucleotideInsertions
    - /aminoAcidInsertions
- other routes should be ignored
- when a new data version is available the cache should be invalidated

fengelniederhammer:
Also make sure to invalidate the cache when SILO has a new data version

resolves #137

## PR Checklist

### DIFF
--- a/lapis2/src/main/kotlin/org/genspectrum/lapis/scheduler/DataVersionCacheInvalidator.kt
+++ b/lapis2/src/main/kotlin/org/genspectrum/lapis/scheduler/DataVersionCacheInvalidator.kt
@@ -1,0 +1,43 @@
+package org.genspectrum.lapis.scheduler
+
+import mu.KotlinLogging
+import org.genspectrum.lapis.silo.CachedSiloClient
+import org.genspectrum.lapis.silo.SILO_QUERY_CACHE_NAME
+import org.springframework.cache.annotation.CacheEvict
+import org.springframework.scheduling.annotation.Scheduled
+import org.springframework.stereotype.Component
+import java.util.concurrent.TimeUnit
+
+private val log = KotlinLogging.logger {}
+
+@Component
+class DataVersionCacheInvalidator(
+    private val cachedSiloClient: CachedSiloClient,
+    private val cacheClearer: CacheClearer,
+) {
+    private var currentlyCachedDataVersion = "uninitialized"
+
+    @Scheduled(fixedRate = 1, timeUnit = TimeUnit.SECONDS)
+    @Synchronized
+    fun invalidateSiloCache() {
+        log.debug { "checking for data version change" }
+
+        val info = cachedSiloClient.callInfo()
+        if (info.dataVersion != currentlyCachedDataVersion) {
+            log.info {
+                "Invalidating cache, old data version: $currentlyCachedDataVersion, " +
+                    "new data version: ${info.dataVersion}"
+            }
+            cacheClearer.clearCache()
+            currentlyCachedDataVersion = info.dataVersion
+        }
+    }
+}
+
+@Component
+class CacheClearer {
+    @CacheEvict(SILO_QUERY_CACHE_NAME, allEntries = true)
+    fun clearCache() {
+        log.info { "Clearing cache $SILO_QUERY_CACHE_NAME" }
+    }
+}
--- a/lapis2/src/main/resources/application.properties
+++ b/lapis2/src/main/resources/application.properties
@@ -2,3 +2,12 @@ springdoc.default-produces-media-type=application/json
 springdoc.api-docs.path=/api-docs
 springdoc.swagger-ui.operationsSorter=alpha
 server.forward-headers-strategy=framework
+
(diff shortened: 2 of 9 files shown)


lapis2/src/main/kotlin/org/genspectrum/lapis/scheduler/DataVersionCacheInvalidator.kt
```diff
@@ -0,0 +1,44 @@
...
+
+@Component
+class DataVersionCacheInvalidator(
+    private val cachedSiloClient: CachedSiloClient,
+    private val cacheClearer: CacheClearer,
+) {
+    final var currentlyCachedDataVersion = "uninitialized"
+        private set
+
+    @Scheduled(fixedRate = 30, timeUnit = TimeUnit.SECONDS)
```
fengelniederhammer, line 21: @chaoran-chen What's a reasonable interval to check for new versions? I think the old LAPIS does it every second, if I interpret this correctly:
https://github.com/GenSpectrum/LAPIS/blob/831399692fdf1bcce6fc9e4021e950c4ee8bc00f/server/src/main/java/ch/ethz/lapis/api/DataVersionService.java#L40

lapis2/src/main/resources/application.properties
```diff
@@ -2,3 +2,12 @@ springdoc.default-produces-media-type=application/json
 springdoc.api-docs.path=/api-docs
 springdoc.swagger-ui.operationsSorter=alpha
 server.forward-headers-strategy=framework
+
+spring.cache.cache-names=siloQueryCache
+spring.cache.caffeine.spec=maximumSize=50000,expireAfterAccess=1h
```
fengelniederhammer, line 7: @chaoran-chen Do you have any experience from LAPIS 1 what reasonable values might be here? As a wild guess, I put 50,000 cached entries at max and a lifetime of 1 hours here.

lapis2/src/main/kotlin/org/genspectrum/lapis/scheduler/DataVersionCacheInvalidator.kt
chaoran-chen: Yes, every second would be good.

lapis2/src/main/resources/application.properties
chaoran-chen: I don't have a good number for the number of entries but 50k sounds reasonable. Alternatively, is there a way to restrict the memory size? For LAPIS 1, the max size was set to 2 GB.

I don't think that we need an expiry time. As long as it's the same data version, the results will be the same.

lapis2/src/main/kotlin/org/genspectrum/lapis/scheduler/DataVersionCacheInvalidator.kt
fengelniederhammer: Ok, then I can probably also remove the other logic that checks the data version on a normal query.

chaoran-chen:
@fengelniederhammer, it just came to my mind (in case you haven't considered it already): if the query asks for a random order, the result may not be cached.

lapis2/src/main/resources/application.properties
fengelniederhammer: Thinking about it: the larget collection page (that I know of) sends around 2200 requests. Normal variant pages send around 20 relevant requests. So 5000 entries should probably do it here.

fengelniederhammer:
Thanks for the reminder! I added a comment to #658

lapis2/src/main/resources/application.properties
chaoran-chen: Why only 5000? Or do you mean 50,000?

Don't forget that a collection is not static. You can filter by location and time range:

image

fengelniederhammer: Ah true. Ok, then I'll put it to 50k and no expiry time. Anyway, this config value can be overwritten when starting LAPIS.

fengelniederhammer: I did not find a way to make the eviction based on memory size. https://github.com/ben-manes/caffeine/wiki/Eviction#size-based

Maybe there are other cache implementations that support it. In principle, it should be easy to exchange. None of the implementation code is specific to the implementation.

chaoran-chen: sounds good, then let's try with 50k and no expiry time

JonasKellerer (commented):
Can we somehow test whether the cache is invalidated when a new data version is available?

Also do we need to add some docu?

fengelniederhammer:
There is `SiloClientAndCacheInvalidatorTest` in `SiloClientTest.kt` that checks tests this: populate the cache with some data version, set the SILO mock to return a new data version and check that the next time, the cache isn't hit.

I'll do the docs in a separate PR.

PR #724, opened 2024-04-02
fix: write empty string to CSV/TSV for null values

### LINKED ISSUE #708: Return emptiness instead of `null` in TSV/CSVs
@corneliusroemer found that in TSV files, the string `null` is provided if a value is `null`. That's only right for JSON (where we really have a `null` value) but in CSVs and TSVs, if a value is unknown, the field should simply be left empty.

fengelniederhammer:
resolves #708

## PR Checklist


JonasKellerer (approved):
LGTM

PR #727, opened 2024-04-03
download compression is too slow

### LINKED ISSUE #717: Download compression is too slow
@corneliusroemer posted in a downstream project:

> When downloading all Ebola data, I was surprised to find that zstd compressed download was very slow.
> 
> The bottleneck is clearly compression on the server, as uncompressed download happened in 2 seconds for me (60 MB raw), whereas zstd compressed download took 8 seconds (60 MB raw -> 340kB compressed).
> 
> That's a meager compression rate (of uncompressed input) of ~8MB/s.
> 
> Running zstd on the CLI on the same uncompresed file gets me a throughput of 2GiB/s and results in a higher compression ratio (327 KiB vs 340 KiB for the server compressed file).
> 
> Clearly the zstd compression on the server can benefit from some tuning.
> 
> Gzip is also slower than it need be, running at only 5MB/s, when on my CLI I get 230MiB/s.

fengelniederhammer:
resolves #717

Tests on a medium-sized data set (3552 sequences):
I measured request times of /sample/unalignedNucleotideSequences:
- uncompressed: 1300 ms, 65 MB
- gzip compressed: 3100 ms (before: 14 s), 8 MB
- zstd compressed: 970 ms (before: 7.4 s), 342 kB

I tested this using the Ebola dataset from Loculus. I pushed it to a branch: https://github.com/GenSpectrum/LAPIS/tree/ebolaTestData

## PR Checklist

### DIFF
--- a/lapis2/src/main/kotlin/org/genspectrum/lapis/controller/CompressionFilter.kt
+++ b/lapis2/src/main/kotlin/org/genspectrum/lapis/controller/CompressionFilter.kt
@@ -222,6 +222,18 @@ class CompressingServletOutputStream(
         compressingStream.write(byte)
     }
 
+    override fun write(bytes: ByteArray) {
+        compressingStream.write(bytes)
+    }
+
+    override fun write(
+        bytes: ByteArray,
+        offset: Int,
+        length: Int,
+    ) {
+        compressingStream.write(bytes, offset, length)
+    }
+
     override fun isReady() = outputStream.isReady
 
     override fun setWriteListener(listener: WriteListener?) = outputStream.setWriteListener(listener)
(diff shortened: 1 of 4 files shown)


chaoran-chen (commented):
Cool! What was the issue before? (I don't really understand what the changed code is doing.)

lapis2/src/main/kotlin/org/genspectrum/lapis/controller/CompressionFilter.kt
```diff
@@ -222,6 +222,18 @@ class CompressingServletOutputStream(
...
+        compressingStream.write(bytes)
+    }
+
+    override fun write(
+        bytes: ByteArray,
+        offset: Int,
+        length: Int,
+    ) {
+        compressingStream.write(bytes, offset, length)
+    }
```
fengelniederhammer, line 235: @chaoran-chen Those are the relevant changes. We didn't forward those methods to the compressing stream. Instead, the default implementation would call `fun write(byte: Int)` for every entry of the `ByteArray`. Apparently the compression streams can do a lot better than writing every byte individually.

JonasKellerer (approved):
LGTM (works on my machine :) )

PR #739, opened 2024-04-17
fix(lapis2): prefer zstd over gzip

### LINKED ISSUE #738: ENH: Offer zstd encoding for pareto superior compression
Recently, zstd has become supported in Chromium alongside br, gzip and deflate as a supported compression method.

It might be worth looking into using it for on the fly compression for LAPIS as that's exactly the domain where zstd is absolutely amazing: fast compression on server side, good compression ratios, fast decompression in browser.

I had that idea when I noticed you use gzip for the many requests for covSpectrum:
image

image

https://caniuse.com/mdn-http_headers_content-encoding_zstd

fengelniederhammer:
resolves #738

Chrome now supports zstd compression, so it will set `Accept-Encoding: gzip, deflate, br, zstd`. In that case, we should prefer zstd, because it's usually faster and has better compression ratios.

## PR Checklist

### DIFF
--- a/lapis2/src/main/kotlin/org/genspectrum/lapis/controller/CompressionFilter.kt
+++ b/lapis2/src/main/kotlin/org/genspectrum/lapis/controller/CompressionFilter.kt
@@ -65,8 +65,8 @@ enum class Compression(
                 .map { it.trim() }
 
             return when {
-                headersList.contains(GZIP.value) -> GZIP
                 headersList.contains(ZSTD.value) -> ZSTD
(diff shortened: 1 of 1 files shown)


corneliusroemer:
I might have been wrong about it having higher compression ratios - I think they are pretty similar and sometimes worse for zstd, but the extra speed and reduced CPU load is worth losing something like a 5% increase in size (I think). People who want max compression can always ask for gzip only.

JonasKellerer (approved):
LGTM (My chrome loads the data)

PR #743, opened 2024-04-18
build(root): use commitlint to enforce conventional commits

### LINKED ISSUE #169: Conventional commits linter
To have even more linters, we could run [commitlint](https://github.com/conventional-changelog/commitlint) in the CI to ensure that the commit messages follow the conventional commit standard.

This does not only apply to this repository but to all of our repositories.

fengelniederhammer:
resolves #169

## PR Checklist

### DIFF
--- a/.commitlintrc.js
+++ b/.commitlintrc.js
@@ -1,0 +1,25 @@
+import {RuleConfigSeverity} from "@commitlint/types";
+
+/**
+ * @type {import('@commitlint/types').UserConfig}
+ */
+const Configuration = {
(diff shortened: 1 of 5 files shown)


pflanze:
The counterpoint with all the linters is that practically I need to run them locally, too; round-tripping through GitHub only to be stopped by another linter after 20 minutes hampers productivity. So I'd like to only OK this if there are instructions on how to use it, and in the future, it's part of a script or makefile so that the process is automatic locally, too.

JonasKellerer (approved):
LGTM

PR #760, opened 2024-04-27
feat: allow filtering for null

### LINKED ISSUE #757: Filter by `null` values
It should be possible to filter for `null` valued fields.

I just checked on the Loculus Ebola instance:

### For GET requests it seems to work:
`GET /sample/details?accessionVersion=LOC_0001D2F.1&ncbi_is_lab_host=` yields a result.
Taken from this swagger request:
grafik

### For POST it does not

```
POST /sample/details
{
  "accessionVersion": "LOC_0001D2F.1",
  "ncbi_is_lab_host": null
}
``` 

yields no result.

```
POST /sample/details
{
  "accessionVersion": "LOC_0001D2F.1"
}
``` 
yields the same result as the GET request.

### Expected behavior

* Both GET and POST should return the same results.
* For both methods it should be possible to filter by `null`. This should be possible for every sequence filter field that LAPIS offers.

JonasKellerer:
resolves #757

## Description
Allows all filters in GET and POST requests to be null (except pango lineage)

## PR Checklist


JonasKellerer:
Currrently this PR is waiting on SILO to handle null values in float filters.

JonasKellerer:
@fengelniederhammer Do we want to explicitly show in Swagger that null is a valid filter value?

fengelniederhammer:
> @fengelniederhammer Do we want to explicitly show in Swagger that null is a valid filter value?

Would be good, does Swagger have good support for nullable types? Or is it that "not required" means "nullable" in Swagger?

lapis2/src/test/kotlin/org/genspectrum/lapis/model/SiloFilterExpressionMapperTest.kt
fengelniederhammer, line 1: Let's also add a test for a `dateFrom` and `dateTo` with explicitly set `null`s? I think `getAsDate` in the implementation needs some null handling, too.

lapis2/src/test/kotlin/org/genspectrum/lapis/silo/SiloQueryTest.kt
```diff
@@ -361,6 +361,16 @@ class SiloQueryTest {
...
                 """,
                 ),
+                Arguments.of(
+                    StringEquals("theColumn", null),
+                    """
+                {
+                    "type": "StringEquals",
+                    "column": "theColumn",
+                    "value": null
+                }
```
fengelniederhammer, line 371:
```suggestion
                    {
                        "type": "StringEquals",
                        "column": "theColumn",
                        "value": null
                    }
```

fengelniederhammer: (Actually also the indent of most other JSONs here is wrong - at least for my taste)

lapis2/src/main/kotlin/org/genspectrum/lapis/model/SiloFilterExpressionMapper.kt
fengelniederhammer, line 1: Line 243 shows a warning:
```
Type mismatch.
Required: CharSequence
Found: String?
```

fengelniederhammer (changes requested):
Looks good :+1:

lapis2/src/main/kotlin/org/genspectrum/lapis/model/SiloFilterExpressionMapper.kt
JonasKellerer: Fixed with null handling.

lapis2/src/main/kotlin/org/genspectrum/lapis/model/SiloFilterExpressionMapper.kt
```diff
@@ -308,7 +322,7 @@ class SiloFilterExpressionMapper(
         val value = extractSingleFilterValue(values, originalKey)
 
         try {
-            return value.toInt()
+            return value?.toInt()
```
fengelniederhammer, line 325: Does this also need a "is blank" handling?
`/sample/aggregated?ageFrom=` returns a bad request with `"ageFrom '' is not a valid integer: For input string: \"\""`, while a POST with `{ "age": null}` is ok.

fengelniederhammer: (Same for floats)

fengelniederhammer (approved):

PR #761, opened 2024-04-29
support form encoded post requests

### LINKED ISSUE #754: Support form encoded POST requests
Support accepting form encoded requests with the content type `application/x-www-form-urlencoded`. This improves browser support, because a browser can simply submit a form like this to get a valid LAPIS call:
```html
<form action="my.lapis/sample/aggregated" method="POST">
    <input type="text" name="country" />
    <input type="submit" value="Submit" />
</form>
``` 
For GET methods, this is already supported by LAPIS, since the form data just ends up in the query string. URLs have a length limited, so this doesn't work for large requests (e.g. someone requesting data for 1000 primary keys).

* This should be supported for every endpoint
* It should support setting the `dataFormat` to return the data as JSON, CSV or TSV.

Implementation-wise we could implement it in two different ways: 
1. accept form value `query` that is expected to contain a JSON that translates to a valid POST request to the respective endpoint.
2. accept all fields separately in the URL encoded form format.

We could also support both. Let's see how well Spring supports form url encoded requests and what is easiest to implement.

fengelniederhammer:
resolves #754

## PR Checklist
- ~~[ ] All necessary documentation has been adapted.~~ -> #765 


lapis2/index.html
JonasKellerer, line 1: Do we keep this file in an example folder?

JonasKellerer: Or use it in some test? Maybe e2e with playwrigth?

lapis2/src/test/kotlin/org/genspectrum/lapis/controller/MultiSegmentedSequenceControllerTest.kt
```diff
@@ -180,49 +145,73 @@ class MultiSegmentedSequenceControllerTest(
...
+            ),
+            Arguments.of(
+                "POST JSON",
+                { route: String ->
+                    postSample(route)
+                        .content("""{}""")
+                        .contentType(MediaType.APPLICATION_JSON)
+                },
+            ),
+            // Spring doesn't support empty form encoded requests
```
JonasKellerer, line 214: Do we need to document this somewhere?

lapis2/src/main/kotlin/org/genspectrum/lapis/util/TryToGuessTheType.kt
```diff
@@ -0,0 +1,41 @@
...
+import com.fasterxml.jackson.databind.node.JsonNodeType.ARRAY
+import com.fasterxml.jackson.databind.node.JsonNodeType.BOOLEAN
+import com.fasterxml.jackson.databind.node.JsonNodeType.NULL
+import com.fasterxml.jackson.databind.node.JsonNodeType.NUMBER
+import com.fasterxml.jackson.databind.node.JsonNodeType.STRING
+import com.fasterxml.jackson.databind.node.LongNode
+import com.fasterxml.jackson.databind.node.TextNode
+import org.genspectrum.lapis.controller.SPECIAL_REQUEST_PROPERTY_TYPES
+
+fun tryToGuessTheType(entry: Map.Entry<String, List<String>>): JsonNode {
```
JonasKellerer, line 16: Is this tested somewhere?

lapis2/src/main/kotlin/org/genspectrum/lapis/util/TryToGuessTheType.kt
```diff
@@ -0,0 +1,41 @@
...
+    return when (jsonNodeType) {
+        null, NULL, STRING -> TextNode(value)
+        BOOLEAN -> when (val booleanValue = value.toBooleanStrictOrNull()) {
+            null -> TextNode(value)
+            else -> BooleanNode.valueOf(booleanValue)
+        }
+
+        NUMBER -> value.toLongOrNull()?.let { LongNode(it) }
+            ?: value.toDoubleOrNull()?.let { DoubleNode(it) }
+            ?: TextNode(value)
```
JonasKellerer, line 37: Why dont we use the NumericNode type here? Or is this not necessary?

JonasKellerer (commented):
Nice addition. Seems to work at my machine.

lapis2/src/test/kotlin/org/genspectrum/lapis/controller/MultiSegmentedSequenceControllerTest.kt
fengelniederhammer: Document how? In the form of a test?

lapis2/index.html
fengelniederhammer: I was also thinking about that... I think I would only keep it if we find an easy way to turn this into a test. On the other hand, there is an e2e test that simulates the behavior.

lapis2/src/main/kotlin/org/genspectrum/lapis/util/TryToGuessTheType.kt
fengelniederhammer: implicitly, yes. There are e.g. tests that check valid / invalid `offset`s

lapis2/src/main/kotlin/org/genspectrum/lapis/util/TryToGuessTheType.kt
fengelniederhammer: `NumericNode` is abstract. I think we need to use a concrete subclass.

lapis2/src/test/kotlin/org/genspectrum/lapis/controller/MultiSegmentedSequenceControllerTest.kt
JonasKellerer: I thought more maintainer facing. Eg. in swagger you can send an empty request and get the data. But With the forms we should somehow inform the maintainer that empty requests are not working. Maybe in the lapis-2 docu?

lapis2/src/main/kotlin/org/genspectrum/lapis/util/TryToGuessTheType.kt
JonasKellerer: If it is not to much of a hassle, I would favour an explicit test, since this is a core part. So we know that we dont mess up the types.

lapis2/src/test/kotlin/org/genspectrum/lapis/controller/MultiSegmentedSequenceControllerTest.kt
fengelniederhammer: An empty form is kind of pointless anyway, isn't it? It literally means that there are no values to fill out. 

(Also as a disclaimer: I just tried this from within the Spring tests, not in an e2e scenario).

lapis2/src/main/kotlin/org/genspectrum/lapis/util/TryToGuessTheType.kt
fengelniederhammer: Done

lapis2/src/test/kotlin/org/genspectrum/lapis/controller/MultiSegmentedSequenceControllerTest.kt
JonasKellerer: As discussed: sending an empty form without input elements attached to it does not make sense.

JonasKellerer (approved):
LGTM

PR #767, opened 2024-05-03
chore(lapis2): add commit tags to images

### LINKED ISSUE #678: Versioned releases or docker tags
It would be cool if LAPIS had some sort of versioned releases, particularly with respect to the Docker images, so one can pin to a particular version. At minimum this could just be tagging images with commit hashes? Or it could be real versioning.

Thanks!

theosanderson:
resolves #678

This should add commit-hash-based tags to images


fengelniederhammer:
#771 is an alternative to this that creates proper releases.

PR #771, opened 2024-05-06
Create releases with Release Please

### LINKED ISSUE #678: Versioned releases or docker tags

fengelniederhammer:
resolves #678 

#770 has been created with this.

## PR Checklist


fengelniederhammer:
Open for discussion: I implemented it such that LAPIS and the docs always get the same version. I guess that's the desired behavior?

That also implies that LAPIS gets a version bump, if we change the docs.

chaoran-chen:
> That also implies that LAPIS gets a version bump, if we change the docs.

Sounds good to me!

.github/workflows/release-lapis2.yml
```diff
@@ -0,0 +1,70 @@
+name: Release LAPIS 2
+
+on:
+  push:
+  workflow_run:
+    workflows: [LAPIS v2]
```
JonasKellerer, line 6: Do we want to run this also when `LAPIS v2 Docs` runs? So each time any of LAPIS or the docs change?
(This would mean we must also wait below on the LAPIS vs docker image

.github/workflows/release-lapis2.yml
```diff
@@ -0,0 +1,70 @@
...
+          repo-token: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Tag Already Built Images With Release Tags
+        if: ${{ steps.release.outputs.release_created }}
+        run: |
+          MAJOR=${{ steps.release.outputs.major }}
+          MINOR=${{ steps.release.outputs.minor }}
+          PATCH=${{ steps.release.outputs.patch }}
+          TAGS=("$MAJOR.$MINOR" "$MAJOR.$MINOR.$PATCH")
+          # TODO (#add-ticket!!!!) include $MAJOR only for releases >= 1.0
```
JonasKellerer, line 62: Which ticket? :)

JonasKellerer (approved):
LGTM

.github/workflows/lapis2-docs.yml
JonasKellerer, line 1: Reminder: Squash before merge

.github/workflows/release-lapis2.yml
fengelniederhammer: Both, The LAPIS v2 workflow and the Docs workflow will run on main. So it will always be triggered on main anyway.

The last step in this workflow must assume that both Docker images have been built. 
* `workflow_run LAPIS v2` unsures that the LAPIS image is ready
* the `wait-on-check-action` ensures that the Docs image is ready.

PR #785, opened 2024-05-07
Only release a single `package`

fengelniederhammer:
https://github.com/GenSpectrum/LAPIS/pull/782 shows that configuring two separate packages in Release Please is not the correct approach - we don't want separate versions for LAPIS and the docs. If we release "LAPIS", then it should be the LAPIS code and the docs.

Also `releases_created` is always set, but maybe to `false`, which is still truthy, if we don't check for `false` explicitly. `release_created` is only set, if a release was created.

#783 shows the result of this PR.

## PR Checklist


JonasKellerer (approved):
LGTM

PR #793, opened 2024-05-15
fix(lapis2-docs): relative links must not end with a `/`

fengelniederhammer:
Otherwise the following happens:
- Be on page, e.g. /references/filters
- Click on link that ends with slash, e.g. `../concepts/pango-lineage-query/`
- now you are on /concepts/pango-lineage-query/
- clicking another relative link, e.g. ``../concepts/variant-query` will result in wrong target, in this case /concepts/concepts/variant-query

## PR Checklist


JonasKellerer (approved):
LGTM

PR #809, opened 2024-05-22
make maybe mutation queries case insensitive

### LINKED ISSUE #797: Maybe operator is case sensitive (in contrast to SQL and my naive user expectation)
I finally had a use case for the `maybe` operator so tried it out.

But my queries with `maybe(T111C)` returned 0 sequences and unfortunately not even an error. So did `maybe[T111C]` (I was thinking this might be analogous to `[1-of: xxx]`

After a bit of digging, I found that I had to capitalize `MAYBE()`. This is a bit unexpected - might be good to allow queries to be case insentive in general. Either cast everything to upper or lower case.

fengelniederhammer:
resolves #797

* Breaking change: throw "bad request" when there is unmatched variant query parser input
  * Stop logging unmatched parser input to stdout.
* make "maybe" case-insensitive in both, variant queries and normal mutation filters

## PR Checklist


JonasKellerer (approved):
LGTM

PR #822, opened 2024-05-29
chore(dockerfile): use eclipse-temurin instead of deprecated openjdk image

corneliusroemer:
- Use eclipse-temurin:21-alpine as run image instead of deprecated openjdk. See deprecation notice here: https://github.com/docker-library/openjdk/issues/505
- Use entrypoint.sh as entrypoint in dockerfile to allow passing JVM_OPTS via env variable. Mostly copied from private repo where it works flawlessly

### DIFF
--- a/lapis2/.dockerignore
+++ b/lapis2/.dockerignore
@@ -4,3 +4,4 @@
 !gradlew
 !build.gradle
 !settings.gradle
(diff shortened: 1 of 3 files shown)


corneliusroemer:
Close in favor of #823

PR #849, opened 2024-06-28
812 outofmemoryerror causes lapis to stop working without crashing

### LINKED ISSUE #812: OutOfMemoryError causes LAPIS to stop working without crashing
I think that sometimes, LAPIS can get an OutOfMemoryError and stop working but the program doesn't crash so Docker won't restart it. We should ensure that LAPIS terminates with a non-zero exit if it enters a non-working state (or avoid ending up in such a state in the first place).

JonasKellerer:
resolves #812

This adds the possibility to specify the size of the cache. Thus, maintainers have the possibility to set the size appropriate for the machine they are using. If the cache limit is reached, it will remove the oldest entry. 

As you see, it calculates the weight of the request by determining the size in bytes. I'm currently not sure if this has any performance implications (casting to string and then to byteArray).

I tested it locally with a smaller cache size. However, automatic tests are not possible, since we do not provide the info if the request was cached or not to the user. Do we want to?

What are good defaults for the cache size, or do we just leave them out?

Another option would be to use redis. But this would need some setup. How do we cant to proceed?

## PR Checklist


JonasKellerer:
As discussed: maintainers should be able to turn off the cache, or set the size of the cache.

JonasKellerer:
This PR now changed in its implementation:
we now use soft references to store the cache values. The garbage collector should now be possible to clear the cache, if memory is needed. The cache size increases to the the heap size. So the maintainer should provide a limit. When running LAPIS in docker the JVM option `-XX:+UseContainerSupport` should suffice to limit the heap size to the available RAM of the docker container.

JonasKellerer:
I will now adopt the documentation accordingly.

lapis-docs/src/content/docs/maintainer-docs/concepts/caching.mdx
```diff
@@ -7,10 +7,17 @@ import CacheSizeEndpointLink from '../../../../components/CacheSizeEndpointLink.
...
+but don't return much data (measured in size of the response body). Details and FASTA actions are expensive, but return
+potentially large amounts of data and are therefore not cached.
 
-Details and FASTA actions are expensive, but return potentially large amounts of data.
+Internally, we use soft references to allow the JVM garbage collector to remove entries from the cache if the memory is needed.
+However, per default the cache is not guaranteed to have a fixed memory size, increasing with each stored entry.
+The maintainer must ensure that enough memory is available to store the cache entries, or provide a limit to the
+heap size of the JVM (eg. `-Xmx4g`, or `-XX:+UseContainerSupport` when running LAPIS in docker )
+
+You can turn off caching by setting the `spring.cache.type` property to `none` in the `application.properties` file, or
```
fengelniederhammer, line 18: A maintainer will not modify the file, since it's already bundled in the Docker image. I think it's possible to provide a file as a maintainer, but I think in the other docs we say that providing the command line arguments is the way to go. -> Let's also just mention that here?

fengelniederhammer (approved):
LGTM. Let's maybe deploy it on the test server to see whether the new cache config survives CoV Spectrum?

PR #885, opened 2024-08-07
[DO NOT MERGE] feature branch regex

### LINKED ISSUE #750: Wildcard based searches
It would be nice to be able to search strings using wildcards. E.g. if my sequences have a `name` column which might look like `AWESOME-LAB/SEQUENCE_22` it would be cool to be able to search with `AWESOME-LAB*` to find all awesome lab sequences.

Taepper:
resolves #750

This feature branch maps string fields from the API to SILO's StringSearch instead of StringEquals expressions

This is a breaking change as any search for a string field will now also assume the value to be a regex string, whereas before it was just a string that was compared against the column values with equality.

The problem with a permanent solution is that the [current api](https://lapis.cov-spectrum.org/gisaid/v2/swagger-ui/index.html#/lapis-controller/postDetails_1) just expects a pair of `"column": "value"`, where not specification of whether the filtering should be done with equality or regex matching is possible.

Possible solutions:
- Upon short consideration, it is expected that string columns are going to need not both types, but only one type of the filter expressions. E.g. author lists will generally be filtered with regexes, the primary key with equality. This could therefore be configured in the column declarations (i.e. in the database_config)
- Make LAPIS parse the `"value"`. If it is of the form `"regex({})"` / `"LIKE {}"`, we will regex-match it, if `"= {}"` we will compare with equality.
- Make LAPIS parse the `"column"`. If it is of the form `"{column}Match"`, we will regex-match, otherwise we will compare with equality.
- Change the API to take objects instead of one `"value"`

## PR Checklist

### DIFF
--- a/lapis/src/main/kotlin/org/genspectrum/lapis/model/SiloFilterExpressionMapper.kt
+++ b/lapis/src/main/kotlin/org/genspectrum/lapis/model/SiloFilterExpressionMapper.kt
@@ -27,7 +27,7 @@ import org.genspectrum.lapis.silo.NucleotideSymbolEquals
 import org.genspectrum.lapis.silo.Or
 import org.genspectrum.lapis.silo.PangoLineageEquals
 import org.genspectrum.lapis.silo.SiloFilterExpression
-import org.genspectrum.lapis.silo.StringEquals
+import org.genspectrum.lapis.silo.StringSearch
 import org.genspectrum.lapis.silo.True
 import org.springframework.stereotype.Component
 import java.time.LocalDate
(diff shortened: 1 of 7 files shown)


chaoran-chen:
This PR will not be merged as such. We are thinking about how to handle it best. I would suggest adding new parameters for string values which have a suffix that indicates that it is a regex search. The question related to our other discussion is then whether the suffix should be `Regex` or `_regex`.

chaoran-chen:
> It might be safer not to rely on these magic parameter names - you could put the type also in the query value, e.g. authors=regex:^[^\t]*$&country=exact:Germany

This would we a major breaking change, wouldn't it? I think that there should continue to be a simple way for people to filter for a exact value without prepending something.

corneliusroemer:
The new field type could be opt in. Old field type would have simple exact string semantics and be backwards compatible.

You can also make the search type gate able via a second parameter: `authors=[^\t]*&authors_type=regex` or something like that 

But yeah then one can just modify the name directly. I see your points 😀

fengelniederhammer:
superseded by #902 (#897 for overview)

PR #906, opened 2024-08-21
docs(lapis): document string search feature

JonasKellerer:
resolves #900

This PR adds the documentation of the string search feature in LAPIS.

## PR Checklist
- [x] All necessary documentation has been adapted.
~~- [ ] The implemented feature is covered by an appropriate test.~~

PR #911, opened 2024-08-22
feat(lapis): use dot symbol as regex separator #908

### LINKED ISSUE #908: Regex search: Try wether `.` works as separator

JonasKellerer:
resolves #908

This PR changes the delimiter symbol of the string search from the `$` sign to a `.` sign. This symbol is not specially encoded in urls. For me this worked on webkit, curl, firefox and chrome. 

Some tools which use ts, like openapi-genereator-cli, have a problem with the dot, since it can not be used in a variable name. The $ sign can be used there. However, kotlin cant use either of them. So this might not an argument for or against this change.

## PR Checklist
   The documentation needs to be changed, but this needs to wait for #906.

### DIFF
--- a/.github/workflows/changelog-preview.yml
+++ b/.github/workflows/changelog-preview.yml
@@ -2,6 +2,11 @@ name: Changelog Preview
 
 on:
   pull_request:
+    types:
+      - opened
+      - reopened
+      - synchronize
+      - ready_for_review
 
 jobs:
   previewChangelog:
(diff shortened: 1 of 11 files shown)


fengelniederhammer (changes requested):
The code looks good. The docs need to be updated, too, if we decide to merge this PR.

fengelniederhammer (approved):
LGTM

Strictly speaking, this is a breaking change, isn't it? So we should make it version 0.3.0?

PR #913, opened 2024-08-26
feat(lapis): omit sequences with `null` values in unaligned fasta downloads

### LINKED ISSUE #912: LAPIS outputs FASTA files reading `null` for missing segments
@rneher pointed out that when downloading unaligned segments that don't exist from LAPIS, one receives output formatted as:

```
>sequence1
null
```

from LAPIS.

This is not "valid" FASTA and would break many pipelines as `null` would be interpreted as a sequence

fengelniederhammer:
closes #912

## PR Checklist


JonasKellerer (approved):
LGTM

PR #962, opened 2024-09-26
fix(lapis-docs): update link in docs introduction.mdx

anna-parker:
Link in introduction is broken, link should be https://lapis-three.vercel.app/maintainer-docs/tutorials/start-lapis-and-silo:

image


fengelniederhammer:
Thanks for finding this!

But #963 is the better fix, because actually the link linking to that page was slightly broken. (I was wondering, since we have a test that is supposed to catch this and the test was working and green)

PR #979, opened 2024-10-10
Use ndjson for e2e tests

fengelniederhammer:
On my machine the tests didn't work because of https://github.com/GenSpectrum/LAPIS-SILO/issues/607.
We should move away from the TSV/Fasta input format anyway because of https://github.com/GenSpectrum/LAPIS-SILO/issues/562.

## PR Checklist


JonasKellerer (approved):
LGTM
I have not verified each file, but since the tests stay the same I think this is okay.

fengelniederhammer:
I also think it's ok as long as the tests are still green (in case I did a mistake when transforming the files). The exact data doesn't matter so much in these tests.

PR #981, opened 2024-10-10
ci(github-actions): improve messages on the generated changelog previews

fengelniederhammer:
We copied this from the components to SILO a while ago, but not yet to LAPIS. I think it has proven itself as good enough so that we can also use it here.

## PR Checklist


JonasKellerer (approved):
LGTM

PR #983, opened 2024-10-14
feat: make it possible to return sequences as json or ndjson

### LINKED ISSUE #971: make it possible to return sequences as JSON or NDJSON
Currently, the sequences endpoints only return FASTA, which is nice for downloading files. But when working with LAPIS programmatically, a structured response would be better.

Implement getting the sequences as JSON or NDJSON (or both, if easily possible). The default should still be FASTA.

Make sure that the result can still be streamed properly. 
* This should be straight forward for NDJSON.
* Think about how to stream JSON. (Maybe just write `[`, then NDJSON lines separated by `,` and then `]`?)

fengelniederhammer:
resolves #971

## PR Checklist


lapis-docs/src/content/docs/concepts/response-format.mdx
```diff
@@ -11,7 +11,14 @@ A '200' response code indicates a successful request, while any other code signi
...
+[FASTA format](https://en.wikipedia.org/wiki/FASTA_format) by default,
+but can also be requested in JSON or [NDJSON](https://github.com/ndjson/ndjson-spec) format.
+
+:::note
+NDJSON is in particular useful for downloading large datasets, as it allows for streaming data processing.
+You can read line by line without loading the entire file into memory.
+Since every line is a valid JSON object, it is usually easier to handle programmatically than FASTA.
+:::
 
 ## Example
```
JonasKellerer, line 24: I think we need to adapt the pictures for the example. The data format is no longer required, but the reportTo is.

grafik

JonasKellerer (approved):
LGTM. Nice work. :)

I think only the documentation pictures need some update, but this is non-blocking.

lapis-docs/src/content/docs/concepts/response-format.mdx
fengelniederhammer: I updated it :+1:

corneliusroemer:
This might have been a breaking change: https://github.com/loculus-project/loculus/pull/3077#issuecomment-2438356390

lapis-e2e/test/alignedNucleotideSequence.spec.ts
corneliusroemer, line 1: You seem to have replaced all FASTA tests with JSON tests, thereby no longer testing that the FASTA is output correctly and not noticing that the default changed as well.

PR #1006, opened 2024-11-11
mark sections of the docs that are generated from the config

### LINKED ISSUE #928: There should be general LAPIS docs, without database specific fields
It appears every single LAPIS instance should set up its own docs instance? I think this is suboptimal: most content will be duplicated. It'd be nice if the invariant content was on a central LAPIS url, so that not every LAPIS user needs to have mostly identical content out there.

One can then just Google something and use the official docs (for invariant content)

Follow up from what I learned in https://github.com/GenSpectrum/LAPIS/issues/924

fengelniederhammer:
resolves #928

This also fixes pango lineage docs (we can't be sure anymore which lineage fields are pango lineages).
Also, all sections are now always available.

## PR Checklist

### DIFF
--- a/lapis-docs/Dockerfile
+++ b/lapis-docs/Dockerfile
@@ -13,4 +13,6 @@ VOLUME /config
 
 HEALTHCHECK --start-period=60s CMD curl --fail --silent localhost:3000$BASE_URL | grep "Welcome to LAPIS" || exit 1
 
-CMD npm ci && npm run build && npm run preview -- --host 0.0.0.0 --port 3000
+RUN npm ci
+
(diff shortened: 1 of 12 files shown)


JonasKellerer (changes requested):
This is nice to have to see all configuration options with their documentation.
I went through the docs and have found these section, which could benefit from the banner:
  - `/getting-started/generate-your-request` full page
  - `/references/open-api-definition` below the section "Swagger UI"
  - `/references/database-config` full page
  - `/references/reference-genomes` full page

JonasKellerer (approved):
LGTM

PR #1036, opened 2024-12-19
test(lapis-e2e): adapt SILO to 0.5.0 changes

fengelniederhammer:
See https://github.com/GenSpectrum/LAPIS-SILO/pull/633

resolves #

## PR Checklist

### DIFF
--- a/lapis/docker-compose.yml
+++ b/lapis/docker-compose.yml
@@ -18,7 +18,7 @@ services:
     image: ghcr.io/genspectrum/lapis-silo:${SILO_TAG}
     ports:
       - "8091:8081"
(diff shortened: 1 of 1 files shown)


JonasKellerer (approved):
LGTM.

I think we should also update the table in the README with the LAPIS versions. Or have it in mind before we release the new version.

fengelniederhammer:
> I think we should also update the table in the README with the LAPIS versions. Or have it in mind before we release the new version.

I think there is actually no change? SILO 0.5.0 is breaking in itself, but not breaking with respect to compatibility with LAPIS.

PR #1053, opened 2025-01-16
feat(lapis): stream JSON responses

### LINKED ISSUE #1047: /sample/details endpoint: Internal server error due to java.lang.OutOfMemoryError when querying a large dataset
Link to issue: https://github.com/loculus-project/loculus/issues/3511#issuecomment-2585193277

fengelniederhammer:
So that the JSON doesn't pile up in memory, risking OOM errors when downloading large datasets as JSON.

resolves #1047

I tested this with the SILO on staging with open Covid data:
```
    --silo.url=https://gs-staging-1.int.genspectrum.org/open/v2/silo
    --lapis.databaseConfig.path=<database config from https://gs-staging-1.int.genspectrum.org/open/v2/sample/databaseConfig>
    --referenceGenomeFilename=lapis-e2e/testData/singleSegmented/reference_genomes.json
    --server.port=8090
```

On the current LAPIS main, I was able to download details 300,000 sequences with a significant memory footprint (LAPIS needed around 1,7 GB). With this branch, I was able to download the 300,000 sequences with basically no memory impact. I also started downloading 3,000,000 sequences. Since I also didn't see any memory impact after a while, I aborted the download.

## PR Checklist
  - Do we need docs for this? 
  - no new tests (because testing that it actually streams is difficult), but the existing tests provide good coverage.


JonasKellerer (approved):
Nice. I can confirm that the endpoints still work as expected. So I think we can use it and see if it fixes the issue.

PR #1054, opened 2025-01-17
fix(lapis): don't show the single segmented sequence download when there are no nucleotide sequences

fengelniederhammer:
In the context of https://github.com/loculus-project/loculus/issues/3388

This will simply show no endpoint to download nucleotide sequences when there are none in the reference genomes. Without this change, it shows the single segmented endpoints that throw index out of bounds exceptions:
https://github.com/GenSpectrum/LAPIS/blob/c6f625e22074731e12f6cdbcb68344350e65503e/lapis/src/main/kotlin/org/genspectrum/lapis/controller/SingleSegmentedSequenceController.kt#L109

## PR Checklist

### DIFF
--- a/lapis/src/main/kotlin/org/genspectrum/lapis/controller/SingleSegmentedSequenceController.kt
+++ b/lapis/src/main/kotlin/org/genspectrum/lapis/controller/SingleSegmentedSequenceController.kt
@@ -40,11 +40,12 @@ import org.springframework.web.bind.annotation.RequestMapping
 import org.springframework.web.bind.annotation.RequestParam
 import org.springframework.web.bind.annotation.RestController
 
-const val IS_SINGLE_SEGMENT_SEQUENCE_EXPRESSION =
-    "#{'\${$REFERENCE_GENOME_SEGMENTS_APPLICATION_ARG_PREFIX}'.split(',').length == 1}"
+const val SHOW_SINGLE_SEGMENTED_CONTROLLER =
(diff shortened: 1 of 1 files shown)


lapis/src/main/kotlin/org/genspectrum/lapis/controller/SingleSegmentedSequenceController.kt
```diff
@@ -41,7 +41,8 @@ import org.springframework.web.bind.annotation.RequestParam
 import org.springframework.web.bind.annotation.RestController
 
 const val IS_SINGLE_SEGMENT_SEQUENCE_EXPRESSION =
```
JonasKellerer, line 43: Should we rename this variable to something like
```suggestion
const val SHOW_SINGLE_SEGMENTED_CONTROLLER =
```

JonasKellerer (approved):
LGTM

lapis/src/main/kotlin/org/genspectrum/lapis/controller/SingleSegmentedSequenceController.kt
chaoran-chen, line 1: Is this distinguishing aligned and non-aligned sequences? It should be possible to have one of them and disable the other.

fengelniederhammer: No, neither LAPIS nor SILO currently distinguish aligned and non-aligned sequences.

PR #1073, opened 2025-02-06
feat(lapis): expose the lineage definition files used by SILO

### LINKED ISSUE #1034: Expose lineage definition file
Just as we expose the reference genomes and the database config we should also expose the lineage definitions file, if there is one. Let's add an endpoint that serves the file in JSON and YAML format.

Open points:
* Where does LAPIS get the file from?
  * Supply the file to LAPIS and directly expose it (just as for the reference genomes and the database config - the difference is that LAPIS needs them anyway, but it doesn't need the lineage definitions)
  * Add an endpoint to SILO that exposes the file and LAPIS only forwards it.
* What should happen if there is no such file?
* We were thinking about having multiple lineage definition files in the future - do we want to keep in mind that we maybe need to be able to extend the endpoint to handle multiple files?

fengelniederhammer:
resolves #1034

## PR Checklist
  - Do we want to mention this specifically in the LAPIS docs?

### DIFF
--- a/README.md
+++ b/README.md
@@ -51,6 +51,7 @@ Higher versions will also work if they are not specified in the table.
 
 | LAPIS  | SILO   |
 |--------|--------|
+| 0.3.14 | 0.5.3  |
 | 0.3.13 | 0.5.2  |
 | 0.3.7  | 0.3.0  |
 | 0.2.10 | 0.2.14 |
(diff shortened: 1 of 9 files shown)


JonasKellerer (approved):
LGTM

PR #1116, opened 2025-03-13
WIP: Generalized advanced queries

anna-parker:
resolves #

In preparation for creating advanced queries I first started generalizing the current variant queries - I will then later move this to a separate file.

## PR Checklist
- [ ] What happens when Maybe is set for lineages or metadata?

### DIFF
--- a/lapis/src/main/kotlin/org/genspectrum/lapis/model/VariantQueryCustomListener.kt
+++ b/lapis/src/main/kotlin/org/genspectrum/lapis/model/VariantQueryCustomListener.kt
@@ -20,6 +20,7 @@ import org.antlr.v4.runtime.RecognitionException
 import org.antlr.v4.runtime.RuleContext
 import org.antlr.v4.runtime.tree.ParseTreeListener
 import org.genspectrum.lapis.config.ReferenceGenomeSchema
+import org.genspectrum.lapis.config.SequenceFilterFields
 import org.genspectrum.lapis.controller.BadRequestException
 import org.genspectrum.lapis.request.ESCAPED_STOP_CODON
 import org.genspectrum.lapis.request.LAPIS_INSERTION_AMBIGUITY_SYMBOL
@@ -39,11 +40,14 @@ import org.genspectrum.lapis.silo.NucleotideSymbolEquals
 import org.genspectrum.lapis.silo.Or
 import org.genspectrum.lapis.silo.SiloFilterExpression
 import org.genspectrum.lapis.silo.StringEquals
+import java.util.Locale
+import kotlin.collections.ArrayDeque
 
 private val log = KotlinLogging.logger { }
 
 class VariantQueryCustomListener(
     val referenceGenomeSchema: ReferenceGenomeSchema,
+    val allowedSequenceFilterFields: SequenceFilterFields,
 ) : VariantQueryBaseListener(),
     ParseTreeListener {
     private val expressionStack = ArrayDeque<SiloFilterExpression>()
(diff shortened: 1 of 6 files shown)


lapis/src/main/kotlin/org/genspectrum/lapis/model/VariantQueryCustomListener.kt
```diff
@@ -141,8 +343,13 @@ class VariantQueryCustomListener(
         if (ctx == null) {
             return
         }
+        try {
+            referenceGenomeSchema.getGeneFromLowercaseName(ctx.geneOrName().text.lowercase()).name
```
fengelniederhammer, line 347: This is an issue of the existing code: I think `getGeneFromLowercaseName` should be renamed to something like "get gene ignoring case" and hide the `.lowercase()` inside. Then the error message would be better, too.

lapis/src/main/kotlin/org/genspectrum/lapis/model/VariantQueryCustomListener.kt
```diff
@@ -141,8 +343,13 @@ class VariantQueryCustomListener(
...
+        } catch (e: RecognitionException) {
+            throw BadRequestException("Gene not implemented", e)
+        }
```
fengelniederhammer, line 350: Do we need this catch block?

lapis/src/main/kotlin/org/genspectrum/lapis/model/VariantQueryCustomListener.kt
```diff
@@ -53,6 +70,191 @@ class VariantQueryCustomListener(
...
+        field ?: throw BadRequestException("Metadata field $metadataName does not exist", null)
+        when (field.type) {
+            SequenceFilterFieldType.String -> {
+                expressionStack.addLast(StringEquals(metadataName, metadataValue))
+            }
+
+            SequenceFilterFieldType.Lineage -> {
+                val lineage = LineageEquals(metadataName, metadataValue, false)
+                expressionStack.addLast(lineage)
+            }
```
fengelniederhammer, line 163: I think we have to discuss how we get the `includeSublineages = true` case here. To me, it's currently not clear. The normal code does this:
https://github.com/GenSpectrum/LAPIS/blob/be1bb873aa0358f4d74eb69be3e08c913d445302/lapis/src/main/kotlin/org/genspectrum/lapis/model/SiloFilterExpressionMapper.kt#L282-L298
But it assumes that every lineage system uses this notation which is probably not true? For now, we could probably also use that logic here.

(Sidenote nitpick: Especially for boolean arguments I find the named arguments syntax really helpful)

fengelniederhammer (commented):
Good start :+1: Now it would be interesting to see what it looks like with separate grammars.

lapis/src/main/kotlin/org/genspectrum/lapis/model/VariantQueryCustomListener.kt
anna-parker: oh yes! thank you for mentioning this, I meant to ask and forgot @chaoran-chen and @fengelniederhammer should the generalized advanced queries include lineages? Will `*` always be used to specify the inclusion of sublineages? Should we add the handling of aliases or will SILO do this?

lapis/src/main/kotlin/org/genspectrum/lapis/model/VariantQueryCustomListener.kt
anna-parker: We need to catch if the gene is invalid - but perhaps `getGeneFromLowercaseName` will also handle this - in any case this error message needs to be improved!

lapis/src/main/kotlin/org/genspectrum/lapis/model/VariantQueryCustomListener.kt
fengelniederhammer: I think for now it's probably good to assume that `*` means "including sublineages". But we should think about that separately again.

Handling of aliases can be left to SILO.

chaoran-chen: I agree that we can assume a trailing `*` to mean including sublineages for lineage fields.

lapis/src/main/kotlin/org/genspectrum/lapis/model/VariantQueryCustomListener.kt
anna-parker: I will make a small PR for this change

anna-parker:
Closing in favor of https://github.com/GenSpectrum/LAPIS/pull/1144

PR #1128, opened 2025-03-21
fix(lapis-docs): mention how arrays must be supplied in form urlencoded requests

### LINKED ISSUE #1089: Inconsistency between how `fields` can be specified for GET and form-data POST requests
For `GET` requests you can make a request:

`https://lapis-main.loculus.org/cchf/sample/details?fields=accession,accessionVersion`

However the corresponding POST request, which is what the Swagger UI generates, doesn't work:

```
curl -X 'POST' \
  'https://lapis-main.loculus.org/cchf/sample/details' \
  -H 'accept: application/json' \
  -H 'Content-Type: application/x-www-form-urlencoded' \
  -d 'fields=accession,accessionVersion'`
```

But one can work around this with:
```
curl -X 'POST' \
  'https://lapis-main.loculus.org/cchf/sample/details' \
  -H 'accept: application/json' \
  -H 'Content-Type: application/x-www-form-urlencoded' \
  -d 'fields=accession&fields=accessionVersion'
```

fengelniederhammer:
I also did some more research. I's a bit unclear how arrays should actually be supplied, some sources also mentioned `fields=value1,value2`. Spring itself does not accept this, so I guess it's ok that we also don't accept it.

resolves #1089

## PR Checklist

### DIFF
--- a/.idea/runConfigurations/LapisOpen.xml
+++ b/.idea/runConfigurations/LapisOpen.xml
@@ -1,6 +1,6 @@
 <component name="ProjectRunConfigurationManager">
   <configuration default="false" name="LapisOpen" type="SpringBootApplicationConfigurationType" factoryName="Spring Boot">
-    <option name="ALTERNATIVE_JRE_PATH" value="corretto-21" />
+    <option name="ALTERNATIVE_JRE_PATH" value="21" />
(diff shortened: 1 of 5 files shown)


JonasKellerer (approved):
LGTM.

PR #1138, opened 2025-03-25
fix(lapis): also return CSV header when there are no data

### LINKED ISSUE #1090: Return CSV/TSV headers even when there are no data
@theosanderson discovered that apparently LAPIS doesn't return anything if the response is empty (for CSV/TSV format). It should at least print the headers.

fengelniederhammer:
resolves #1090

## PR Checklist

### DIFF
--- a/lapis/src/main/kotlin/org/genspectrum/lapis/controller/LapisController.kt
+++ b/lapis/src/main/kotlin/org/genspectrum/lapis/controller/LapisController.kt
@@ -5,6 +5,7 @@ import io.swagger.v3.oas.annotations.Parameter
 import io.swagger.v3.oas.annotations.media.Schema
 import jakarta.servlet.http.HttpServletRequest
 import jakarta.servlet.http.HttpServletResponse
+import org.genspectrum.lapis.config.DatabaseConfig
 import org.genspectrum.lapis.controller.LapisMediaType.TEXT_CSV_VALUE
 import org.genspectrum.lapis.controller.LapisMediaType.TEXT_TSV_VALUE
 import org.genspectrum.lapis.controller.LapisMediaType.TEXT_X_FASTA_VALUE
@@ -52,9 +53,13 @@ import org.genspectrum.lapis.request.OrderByField
 import org.genspectrum.lapis.request.SPECIAL_REQUEST_PROPERTIES
 import org.genspectrum.lapis.request.SequenceFiltersRequest
 import org.genspectrum.lapis.request.SequenceFiltersRequestWithFields
+import org.genspectrum.lapis.response.AggregatedCollection
 import org.genspectrum.lapis.response.Delimiter.COMMA
 import org.genspectrum.lapis.response.Delimiter.TAB
+import org.genspectrum.lapis.response.DetailsCollection
+import org.genspectrum.lapis.response.InsertionsCollection
 import org.genspectrum.lapis.response.LapisResponseStreamer
+import org.genspectrum.lapis.response.MutationsCollection
 import org.genspectrum.lapis.response.ResponseFormat
 import org.genspectrum.lapis.response.SequencesStreamer
 import org.genspectrum.lapis.silo.SequenceType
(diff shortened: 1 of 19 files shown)


lapis/src/main/kotlin/org/genspectrum/lapis/response/CsvWriter.kt
```diff
@@ -1,59 +1,47 @@
...
-    @JsonIgnore
-    fun getHeader(comparator: Comparator<String>): List<String>
+    fun getHeader(): List<String>
+
+    fun getCsvRecords(): Stream<List<String?>> = records.map { mapToCsvValuesList(it) }
+
+    /**
+     * Csv values - must be in the same order as the header.
+     */
+    fun mapToCsvValuesList(value: T): List<String?>
```
JonasKellerer, line 18: Is it possible to test this somehow for our specific collections in `ResponseCollections.kt`? Just to make sure that when we change things with the collections in the future, this will be still true.

lapis/src/main/kotlin/org/genspectrum/lapis/controller/LapisController.kt
```diff
@@ -1069,6 +1091,18 @@ class LapisController(
...
 
+    private fun getDetailsCollection(request: SequenceFiltersRequestWithFields): DetailsCollection {
+        val fields = request.fields.map { it.fieldName }
+
+        return DetailsCollection(
+            records = siloQueryModel.getDetails(request),
+            fields = when (fields.isEmpty()) {
+                true -> databaseConfig.schema.metadata.map { it.name }
+                false -> fields
+            },
```
JonasKellerer, line 1102: Good idea to put the sorting right here.

lapis/src/test/kotlin/org/genspectrum/lapis/controller/Helpers.kt
```diff
@@ -53,3 +54,15 @@ fun sequenceFiltersRequestWithArrayValuedFilters(
...
     emptyList(),
 )
+
+fun MockHttpServletRequestBuilder.withFieldsQuery(fields: List<String>?) =
+    fields?.fold(this) { request, field -> request.queryParam("fields", field) } ?: this
+
+fun MockHttpServletRequestBuilder.withFieldsParam(fields: List<String>?) =
+    fields?.fold(this) { request, field -> request.param("fields", field) } ?: this
+
+fun getFieldsJsonPart(fields: List<String>?) =
```
JonasKellerer, line 64: Nitpick: I would name this `getFieldsAsJsonPart`.

lapis/src/test/resources/config/testDatabaseConfig.yaml
```diff
@@ -15,6 +15,10 @@ schema:
       generateLineageIndex: true
     - name: test_boolean_column
       type: boolean
+    - name: age
```
JonasKellerer, line 18: Nitpick: Should we name the int column also like the other columns: `test_int_column`, and not `age`. 
Also maybe not this PR, but we could also rename the other columns to fit the same format.

JonasKellerer (approved):
Nice work. There was more refactoring going on than I would have expected, but it looks good. 
Only some minor issues.

lapis/src/main/kotlin/org/genspectrum/lapis/response/ResponseCollections.kt
```diff
@@ -0,0 +1,84 @@
...
+class MutationsCollection(
+    override val records: Stream<MutationResponse>,
+) : RecordCollection<MutationResponse> {
+    override fun getHeader() =
+        listOf(
+            "mutation",
+            "count",
+            "coverage",
+            "proportion",
+            "sequenceName",
```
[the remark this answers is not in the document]
fengelniederhammer: I think "sequenceName" is a good generic term for both, gene and segment. Anyway that would be a breaking change, so I'll ignore this here.

lapis/src/main/kotlin/org/genspectrum/lapis/response/CsvWriter.kt
fengelniederhammer: The tests that check the CSVs of insertions and mutations implicitly test that already.

lapis/src/test/resources/config/testDatabaseConfig.yaml
fengelniederhammer: Sounds like a good idea, but I'll leave that to a separate PR, if someone has time.

PR #1149, opened 2025-04-08
feat(lapis): add `fields` property to mutation endpoints

### LINKED ISSUE #1092: Think about: `fields` on mutations endpoints?
https://github.com/GenSpectrum/LAPIS-SILO/issues/711 adds the possibility to filter for `fields` that the response of a mutations action should contain. Do we also want to expose this option in LAPIS?

fengelniederhammer:
If `fields` is not empty, then the response will only contain the fields that were specified there.

resolves #1092

## PR Checklist


JonasKellerer (approved):
Great work. Only some minor refactoring and one unit test case left.

PR #1156, opened 2025-04-15
test(lapis-e2e): fix mutation e2e tests

fengelniederhammer:
I think this was introduced by first merging #1149 (which added the tests) and then merging a non-rebased #1135 (which changed the open api generator version).

resolves #

## PR Checklist


anna-parker:
Can we add these tests to the .github file so that we don't merge tests with an error again?

anna-parker (approved):
lgtm

Issue #1162, opened 2025-04-24
e2e tests are unsorted


anna-parker:
I realized while testing that multiple e2e tests assume a consistent output order of results even when no order is specified - after adding advancedQueries this does not always hold.

Current results on main for nucleotideInsertions e2e:
```
{
  "data": [
    {
      "insertion": "ins_22204:CAGAA",
      "count": 1,
      "insertedSymbols": "CAGAA",
      "position": 22204,
      "sequenceName": null
    },
    {
      "insertion": "ins_22339:GCTGGT",
      "count": 1,
      "insertedSymbols": "GCTGGT",
      "position": 22339,
      "sequenceName": null
    },
    {
      "insertion": "ins_5959:TAT",
      "count": 2,
      "insertedSymbols": "TAT",
      "position": 5959,
      "sequenceName": null
    },
    {
      "insertion": "ins_25701:CCC",
      "count": 17,
      "insertedSymbols": "CCC",
      "position": 25701,
      "sequenceName": null
    }
  ],
  "info": {
    "dataVersion": "1745483239",
    "requestId": "b8f10539-5ea0-40dd-b7e2-cc11505da088",
    "requestInfo": "sars_cov-2_minimal_test_config on localhost at 2025-04-24T08:28:31.570540180",
    "reportTo": "Please report to https://github.com/GenSpectrum/LAPIS/issues in case you encounter any unexpected issues. Please include the request ID and the requestInfo in your report.",
    "lapisVersion": "547c9c2e4a2ce2e15c64e6de1f155ded31b4934d",
    "siloVersion": "0.6.0"
  }
}
```

PR #1163, opened 2025-04-24
fix(lapis-e2e): remove tests that rely on order without specifying lapis to sort output

anna-parker:
resolves https://github.com/GenSpectrum/LAPIS/issues/1162

## PR Checklist

### DIFF
--- a/lapis-e2e/test/alignedNucleotideSequence.spec.ts
+++ b/lapis-e2e/test/alignedNucleotideSequence.spec.ts
@@ -27,8 +27,10 @@ describe('The /alignedNucleotideSequence endpoint', () => {
     });
 
     expect(result).to.have.length(6);
-    expect(result[0].primaryKey).to.equal('key_5');
-    expect(result[0].m).to.equal('TGGG');
+    const match = result.find(
+      (item: { primaryKey: string; m: string | any[] }) => item.primaryKey === 'key_5' && item.m === 'TGGG'
+    );
+    expect(match).to.exist;
   });
 
   it('should order ascending by specified fields', async () => {
--- a/lapis-e2e/test/alignedNucleotideSequence.spec.ts
+++ b/lapis-e2e/test/alignedNucleotideSequence.spec.ts
@@ -19,7 +19,7 @@ describe('The /alignedNucleotideSequence endpoint', () => {
       return 0;
     });
     expect(result[0].primaryKey).to.equal('key_1001493');
-    expect(result[0].main).to.have.length(59806);
+    expect(result[0].main).to.have.length(29903);
   });
 
   it('should return aligned nucleotide sequences for multi segmented sequences', async () => {
--- a/lapis-e2e/test/alignedNucleotideSequence.spec.ts
+++ b/lapis-e2e/test/alignedNucleotideSequence.spec.ts
@@ -4,7 +4,6 @@ import {
   expectIsZstdEncoded,
   lapisMultiSegmentedSequenceController,
   lapisSingleSegmentedSequenceController,
-  sequenceData,
 } from './common';
 
 describe('The /alignedNucleotideSequence endpoint', () => {
@@ -14,8 +13,11 @@ describe('The /alignedNucleotideSequence endpoint', () => {
     });
 
     expect(result).to.have.length(100);
-    expect(result[0].primaryKey).to.equal('key_3259931');
-    expect(result[0].main).to.have.length(29903);
+    const match = result.find(
+      (item: { primaryKey: string; main: string | any[] }) =>
+        item.primaryKey === 'key_3259931' && item.main?.length === 29903
+    );
+    expect(match).to.exist;
   });
 
   it('should return aligned nucleotide sequences for multi segmented sequences', async () => {
--- a/lapis-e2e/test/alignedNucleotideSequence.spec.ts
+++ b/lapis-e2e/test/alignedNucleotideSequence.spec.ts
@@ -13,7 +13,7 @@ describe('The /alignedNucleotideSequence endpoint', () => {
     });
 
     expect(result).to.have.length(100);
-    result.data.sort((a: any, b: any) => {
+    result.sort((a: any, b: any) => {
       if (a.primaryKey < b.primaryKey) return -1;
       if (a.primaryKey > b.primaryKey) return 1;
       return 0;
@@ -29,7 +29,7 @@ describe('The /alignedNucleotideSequence endpoint', () => {
     });
 
     expect(result).to.have.length(6);
-    result.data.sort((a: any, b: any) => {
+    result.sort((a: any, b: any) => {
       if (a.primaryKey < b.primaryKey) return -1;
       if (a.primaryKey > b.primaryKey) return 1;
       return 0;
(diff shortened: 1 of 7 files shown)


lapis-e2e/test/alignedNucleotideSequence.spec.ts
```diff
@@ -14,8 +13,11 @@ describe('The /alignedNucleotideSequence endpoint', () => {
     });
 
     expect(result).to.have.length(100);
-    expect(result[0].primaryKey).to.equal('key_3259931');
-    expect(result[0].main).to.have.length(29903);
+    const match = result.find(
+      (item: { primaryKey: string; main: string | any[] }) =>
+        item.primaryKey === 'key_3259931' && item.main?.length === 29903
+    );
+    expect(match).to.exist;
```
fengelniederhammer, line 20: This only yields "AssertionError: expected undefined to exist" when it fails -> I would prefer to `orderBy` the result instead. The tests will be simpler, more stable und yield better error messages.

lapis-e2e/test/details.spec.ts
```diff
@@ -11,7 +11,12 @@ describe('The /details endpoint', () => {
     });
 
     expect(result.data).to.have.length(2);
-    expect(result.data[0]).to.be.deep.equal({
+    result.data.sort((a: any, b: any) => {
+      if (a.division < b.division) return -1;
+      if (a.division > b.division) return 1;
+      return 0;
+    });
```
fengelniederhammer, line 18: Let's rather sort using LAPIS?

fengelniederhammer (changes requested):

lapis-e2e/test/alignedNucleotideSequence.spec.ts
```diff
@@ -14,7 +13,12 @@ describe('The /alignedNucleotideSequence endpoint', () => {
     });
 
     expect(result).to.have.length(100);
-    expect(result[0].primaryKey).to.equal('key_3259931');
+    result.sort((a: any, b: any) => {
```
fengelniederhammer, line 16: Explicit `any` is even worse that implicit `any`. It simply kills all type information (whereas having no types leaves it open to be inferred).
```suggestion
    result.sort((a, b) => {
```

lapis-e2e/test/alignedNucleotideSequence.spec.ts
```diff
@@ -14,7 +13,12 @@ describe('The /alignedNucleotideSequence endpoint', () => {
...
+      if (a.primaryKey < b.primaryKey) return -1;
+      if (a.primaryKey > b.primaryKey) return 1;
+      return 0;
```
fengelniederhammer, line 19:
```suggestion
      return a.primaryKey.localeCompare(b.primaryKey)
```

lapis-e2e/test/aminoAcidSequence.spec.ts
```diff
@@ -9,8 +9,11 @@ describe('The /alignedAminoAcidSequence endpoint', () => {
     });
 
     expect(result).to.have.length(100);
-    expect(result[0].primaryKey).to.equal('key_3259931');
-    expect(result[0].s).to.have.length(1274);
+    const match = result.find(
+      (item: { primaryKey: string; s: string | any[] }) =>
+        item.primaryKey === 'key_3259931' && item.s?.length === 1274
+    );
```
fengelniederhammer, line 15: Let's also sort here instead?

lapis-e2e/test/alignedNucleotideSequence.spec.ts
fengelniederhammer: We can also use this in many of the other places.

lapis-e2e/test/nucleotideInsertions.spec.ts
```diff
@@ -54,23 +54,25 @@ describe('The /nucleotideInsertions endpoint', () => {
...
 
     const resultWithLimitAndOffset = await lapisClient.postNucleotideInsertions({
       insertionsRequest: {
-        orderBy: [{ field: 'count', type: 'ascending' }],
-        limit: 2,
+        orderBy: [{ field: 'count', type: 'descending' }],
+        limit: 4,
         offset: 1,
       },
     });
```
fengelniederhammer, line 72: This request is supposed to be the same as the first in this test. IMO reverting order makes it too confusing.

Did you also change it due to the "the first two positions both have count 1 so their order could change" thing? I think the better alternative would be to further specify the order to make it non-ambiguous (e.g. additionally ordering by insertion or similar).

lapis-e2e/test/unalignedNucleotideSequence.spec.ts
```diff
@@ -8,8 +8,11 @@ describe('The /unalignedNucleotideSequence endpoint', () => {
     });
 
     expect(result).to.have.length(100);
-    expect(result[0].primaryKey).to.equal('key_3086369');
-    expect(result[0].main).to.have.length(29903);
+    const match = result.find(
+      (item: { primaryKey: string; main: string | any[] }) =>
+        item.primaryKey === 'key_3086369' && item.main?.length === 29903
+    );
+    expect(match).to.exist;
```
fengelniederhammer, line 15: Same here: Let's also sort instead?

fengelniederhammer (changes requested):

lapis-e2e/test/alignedNucleotideSequence.spec.ts
anna-parker: I need to add a type or I get a linter error

fengelniederhammer: What linter error? I don't think we have a linter in those tests. Then let's add something that's more useful (I also noticed that our general TS setup is not really good here and there are places where tsc complains, but unfortunately we're not running it in the pipeline...)

anna-parker: I get the error: "Parameter 'a' implicitly has an 'any' type" which is a TypeScript error - Specifically, it’s a TypeScript compiler error about missing type annotations.

fengelniederhammer: If you have that here you'll also have it elsewhere in the tests I guess. I'd suggest to ignore it for now, this is not an issue that we need to solve in this PR. The actual problem is somewhere else since `result` already seems to have missing type infos (if `result` had proper types, TS would be able to infer the types of `a` and `b` here).

fengelniederhammer (approved):
LGTM

fengelniederhammer:
I noticed too late - this shouldn't have been a `fix` commit. Now it appears in the changelog, but it shouldn't because it's quite irrelevant to users. Please make sure to pay attention to choosing the correct type.

anna-parker:
> Please make sure to pay attention to choosing the correct type.

so what would this be? a chore? I feel like it should be clear to users as this affects lapis-e2e tests that it isn't relevant. If we are worried about this maybe changes affecting the lapis-e2e tests shouldn't be in the changelog

fengelniederhammer:
> > Please make sure to pay attention to choosing the correct type.
> 
> so what would this be? a chore? I feel like it should be clear to users as this affects lapis-e2e tests that it isn't relevant. If we are worried about this maybe changes affecting the lapis-e2e tests shouldn't be in the changelog

Chore would probably be good, yes. 

I'm not sure whether we can configure release please in such a way that it ignore certain component. In the end it just scans the commit history.

PR #1174, opened 2025-05-05
fix(lapis): fix compiler warning

anna-parker:
resolves #

When building we get a compiler error:
```
Override 'fun getContentLength(str: String, contentType: MediaType?): Long?' has incorrect nullability in its signature compared to the overridden declaration 'fun getContentLength(str: String, contentType: MediaType?): Long'
```

The original method is declared in Java as returning Long (boxed), not long (primitive).

In Java, Long can be null, which aligns with the method docstring: "Returns the content length... or null if not known."

However, Kotlin treats Java’s boxed Long (java.lang.Long) as Long!, which is platform-typed — and when you override such a method, you must match the signature exactly, not make it more nullable.

## PR Checklist

### DIFF
--- a/lapis/src/main/kotlin/org/genspectrum/lapis/controller/middleware/CompressionFilter.kt
+++ b/lapis/src/main/kotlin/org/genspectrum/lapis/controller/middleware/CompressionFilter.kt
@@ -307,6 +307,9 @@ class StringHttpMessageConverterWithUnknownContentLengthInCaseOfCompression(
     environment: Environment,
     private val requestCompression: RequestCompression,
 ) : StringHttpMessageConverter(getCharsetFromEnvironment(environment)) {
+    // The original method is declared in Java as returning Long (can be null)
+    // but in Kotlin it is Long! (platform type) which is not nullable.
+    @Suppress("INCOMPATIBLE_OVERRIDE")
     override fun getContentLength(
         str: String,
         contentType: MediaType?,


fengelniederhammer (changes requested):
The current implementation is correct, it behaves exactly as the documentation says: `@return the content length, or {@code null} if not known`
https://github.com/spring-projects/spring-framework/blob/d8b506e09a475638cfd9ae3f821e9f5378296b8d/spring-web/src/main/java/org/springframework/http/converter/AbstractHttpMessageConverter.java#L292

The method is used here:
https://github.com/spring-projects/spring-framework/blob/d8b506e09a475638cfd9ae3f821e9f5378296b8d/spring-web/src/main/java/org/springframework/http/converter/AbstractHttpMessageConverter.java#L266

Which then calls `HttpHeaders::setContentLength`, which throws when it gets a negative number.
https://github.com/spring-projects/spring-framework/blob/d8b506e09a475638cfd9ae3f821e9f5378296b8d/spring-web/src/main/java/org/springframework/http/HttpHeaders.java#L1043-L1048

I'm not 100% sure whether we will need this custom converter, but this change is too dangerous. The Kotlin warning is unfortunate, but it's a false positive.

anna-parker:
Is there a way to at least ignore the error then?

Otherwise we should add a comment

fengelniederhammer:
Adding a comment sounds good. I think I checked whether we can ignore it and didn't find anything, but I'm not sure. Maybe I also just didn't search thoroughly enough.

fengelniederhammer (approved):
Nice, thanks!

PR #1179, opened 2025-05-06
refactor(lapis): it's cleaner to hand out `null` when getting an unknown sequence from the reference genome

fengelniederhammer:
We don't know yet whether we want to handle that as a "bad request" at this point.

## PR Checklist

### DIFF
--- a/lapis/src/main/kotlin/org/genspectrum/lapis/config/ReferenceGenome.kt
+++ b/lapis/src/main/kotlin/org/genspectrum/lapis/config/ReferenceGenome.kt
@@ -3,7 +3,6 @@ package org.genspectrum.lapis.config
 import com.fasterxml.jackson.annotation.JsonIgnoreProperties
 import com.fasterxml.jackson.module.kotlin.jacksonObjectMapper
 import com.fasterxml.jackson.module.kotlin.readValue
-import org.genspectrum.lapis.controller.BadRequestException
 import java.io.File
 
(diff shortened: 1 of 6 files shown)


anna-parker (commented):
As this makes the code more verbose I think it might be nice to have the getNucleotideSequence function as you suggest not throw an error but add a getNucleotideSequenceOrError function which can use the code as is currently used

fengelniederhammer:
More verbose code is not necessarily bad. Having the exception explicit helps in understanding. IMO this design would have been the appropriate one form the beginning (since it's essentially a map and maps should not throw when accessing a key that's not present)

But I can also change it.

anna-parker:
I just added a new function:

```
fun hasNucleotideSequence(name: String): Boolean = nucleotideSequenceNames.containsKey(name.lowercase())
```
so we can still keep using the old function with the in-built exception throw

fengelniederhammer:
> I just added a new function:
> 
> ```
> fun hasNucleotideSequence(name: String): Boolean = nucleotideSequenceNames.containsKey(name.lowercase())
> ```
> 
> so we can still keep using the old function with the in-built exception throw

That's also good, but I like this approach more, because it has both in a single method. Users can simply check for "has sequence" by checking whether it is null or not.

fengelniederhammer:
> This makes rebasing my PR really annoying I don't really appreciate just merging this without discussing it first.

Sorry, I didn't really think about that. You could still merge main into your branch if that makes it easier?

Issue #1124, opened 2025-03-19
Docs don't "allow our host" anymore


fengelniederhammer:
![Image](https://github.com/user-attachments/assets/2617ac82-3ec2-4f3d-9018-83d22d39f876)

This probably came due to some dependency update recently. It's hopefully easy to fix.

PR #1180, opened 2025-05-09
feat(lapis-docs): allow docs to be run locally

anna-parker:
resolves #

## PR Checklist

### DIFF
--- a/lapis-docs/astro.config.mjs
+++ b/lapis-docs/astro.config.mjs
@@ -225,6 +225,5 @@ export default defineConfig({
     base: process.env.BASE_URL,
     site: process.env.ASTRO_SITE,
     server: {
-        allowedHosts: true,
     },
 });
--- a/lapis-docs/src/config.ts
+++ b/lapis-docs/src/config.ts
@@ -1,5 +1,7 @@
 import fs from 'fs';
 import { parse } from 'yaml';
+import dotenv from 'dotenv';
+dotenv.config();
 
 export type MetadataType =
     | 'string'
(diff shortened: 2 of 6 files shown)


lapis-docs/astro.config.mjs
```diff
@@ -224,7 +224,5 @@ export default defineConfig({
     },
     base: process.env.BASE_URL,
     site: process.env.ASTRO_SITE,
-    server: {
-        allowedHosts: true,
-    },
+    server: {},
```
fengelniederhammer, line 227: Why did you remove this? We recently added it to resolve https://github.com/GenSpectrum/LAPIS/issues/1124.

lapis-docs/src/config.ts
```diff
@@ -1,5 +1,7 @@
 import fs from 'fs';
 import { parse } from 'yaml';
+import dotenv from 'dotenv';
+dotenv.config();
```
fengelniederhammer, line 4: Why do we need this? Actually vite should already make sure that .env files are loaded.

lapis-docs/astro.config.mjs
anna-parker: The current set up throws an error (see PR description):
```
(node:14371) ExperimentalWarning: Type Stripping is an experimental feature and might change at any time
(Use node --trace-warnings ... to show where the warning was created)
[config] Astro found issue(s) with your configuration:

! server.allowedHosts: Did not match union.
  > Expected type { 0: string } | 
  > Received [ true ]
  ```

lapis-docs/src/config.ts
anna-parker: this didnt work for me (I got variable undefined warnings)and @JonasKellerer had a comment in the code that there was an issue with this - adding dotenv fixed the issue for me locally- however I sadly think this is causing the e2e test failures

PR #1182, opened 2025-05-12
chore(lapis): the antlr context should never be null

fengelniederhammer:
and if it ever is, let's rather accept the null pointer exception rather than silently ignoring user input

## PR Checklist


anna-parker (approved):
I guess this might lead to LAPIS failing but we should see the errors then

PR #1186, opened 2025-05-14
fix(lapis-docs): fix condition when to show the "variant queries are not available" banner

fengelniederhammer:
## PR Checklist


JonasKellerer (approved):
Good catch.

PR #1194, opened 2025-05-20
feat(lapis): allow regex search for all string fields by default

### LINKED ISSUE #1175: remove `lapisAllowsRegexSearch` and allow regex search for all string fields
Currently metadata fields need to be configured with `lapisAllowsRegexSearch: true` to allow regex search. We decided to remove that. All string fields should allow regex search.

fengelniederhammer:
BREAKING CHANGE: `lapisAllowsRegexSearch` on string metadata fields in the database config is not necessary anymore, regex search is always enabled for all string fields.

resolves #1175

## PR Checklist

### DIFF
--- a/.commitlintrc.js
+++ b/.commitlintrc.js
@@ -20,6 +20,7 @@ const Configuration = {
         ],
         "header-max-length": [RuleConfigSeverity.Disabled],
         "body-max-line-length": [RuleConfigSeverity.Disabled],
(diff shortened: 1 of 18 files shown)


fengelniederhammer:
> LGTM. We should then also add a servers ticket, to also remove the flag.

I added https://github.com/GenSpectrum/servers/issues/276

anna-parker (approved):

PR #1198, opened 2025-05-26
fix(lapis): advanced queries: make `.regex` case-insensitive

### LINKED ISSUE #1187: Make advanced queries case insensitive
Advanced queries (#1041) should be fully case insensitive:
* keywords such as `and` should be accepted in any casing: `and`, `AND` and `aNd`.
  * make sure that metadata queries that contain a keyword are parsed correctly, e.g.: `my_and_field = my_or_value`)
* metadata field names should be case insensitive: `country = Germany` and `cOuNTry = Germany
  * The values of the metadata (right hand side of the `=`) queries should still remain case sensitive.
* The `.regex` modifier should also be case insensitive.
* Symbols in mutation and insertion queries should be case insensitive.

fengelniederhammer:
also refactor the test to make it easier to grasp what is already tested

resolves #1187

## PR Checklist

### DIFF
--- a/lapis/src/main/kotlin/org/genspectrum/lapis/model/AdvancedQueryCustomListener.kt
+++ b/lapis/src/main/kotlin/org/genspectrum/lapis/model/AdvancedQueryCustomListener.kt
@@ -1,7 +1,6 @@
 package org.genspectrum.lapis.model
 
 import AdvancedQueryBaseListener
-import AdvancedQueryParser
 import AdvancedQueryParser.AndContext
 import AdvancedQueryParser.MaybeContext
 import AdvancedQueryParser.NOfQueryContext
@@ -42,14 +41,6 @@ import java.time.LocalDate
 import java.time.format.DateTimeParseException
 import java.util.Locale
 import kotlin.collections.ArrayDeque
-import kotlin.collections.Map
-import kotlin.collections.associateBy
-import kotlin.collections.joinToString
-import kotlin.collections.listOf
-import kotlin.collections.mutableListOf
-import kotlin.collections.plus
-import kotlin.collections.plusAssign
-import kotlin.collections.reversed
 
 fun getFieldOrThrow(
     metadataFieldsByName: Map<String, DatabaseMetadata>,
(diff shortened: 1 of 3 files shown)


lapis/src/main/kotlin/org/genspectrum/lapis/model/AdvancedQueryCustomListener.kt
```diff
@@ -41,6 +42,14 @@ import java.time.LocalDate
 import java.time.format.DateTimeParseException
 import java.util.Locale
 import kotlin.collections.ArrayDeque
+import kotlin.collections.Map
```
anna-parker, line 45: Are these imports actually needed? I don't see any code changes in this file that would require them

lapis/src/test/kotlin/org/genspectrum/lapis/model/AdvancedQueryFacadeTest.kt
```diff
@@ -237,26 +221,6 @@ class AdvancedQueryFacadeTest {
...
-    fun `GIVEN a advancedQuery with a mixed-case 'Maybe' expression THEN returns 'Maybe' SiloQuery`() {
-        val advancedQuery = "maYbE(T12C)"
-
-        val result = underTest.map(advancedQuery)
-
-        val expectedResult = Maybe(NucleotideSymbolEquals(null, 12, "C"))
-        assertThat(result, equalTo(expectedResult))
-    }
-
     @Test
```
anna-parker, line 224: I find it odd to move the Maybe and not into the parametrized tests but not do the same for the `or` and `and`

lapis/src/main/kotlin/org/genspectrum/lapis/model/AdvancedQueryCustomListener.kt
fengelniederhammer: That's weird, thanks for catching this. I don't know how they ended up there...

lapis/src/test/kotlin/org/genspectrum/lapis/model/AdvancedQueryFacadeTest.kt
```diff
@@ -743,6 +639,82 @@ class AdvancedQueryFacadeTest {
                     "floatFieldTo=1",
                     "Metadata field floatFieldTo does not exist",
                 ),
+                InvalidTestCase(
```
anna-parker, line 642: this should be added to the regexCases list

anna-parker (approved):
This is quite a refactor but as it doesn't seem to actually change much code I decided to just approve to not block too much - please check if all the additional imports are required though before merging

PR #1212, opened 2025-06-03
feat(lapis): allow advanced queries to be used with simple variant an…

anna-parker:
…d metadata queries

resolves #

## PR Checklist


fengelniederhammer:
But I think it should be a `fix(lapis): ...` commit.

PR #1217, opened 2025-06-04
fix(lapis): fix Swagger UI layout of GET requests

### LINKED ISSUE #1193: Swagger UI: sequenceFilters in GET requests looks unexpected
Currently, it looks like this for all GET endpoints: 

Image

Our current GenSpectrum Loculus still runs LAPIS 0.4.1, there is looks as expected:
Image

fengelniederhammer:
in #1135 we upgraded to org.springdoc:springdoc-openapi-starter-webmvc-ui 2.8.6. Since version 2.8.0, they use OpenAPI version 3.1 as default which was breaking for us, because the generated spec didn't look as expected anymore

resolves #1193

## PR Checklist

### DIFF
--- a/.github/workflows/lapis.yml
+++ b/.github/workflows/lapis.yml
@@ -82,18 +82,6 @@ jobs:
       - name: Setup Gradle
         uses: gradle/actions/setup-gradle@v4
 
-      - name: Build OpenAPI Spec
-        working-directory: lapis
-        run: ./gradlew generateOpenApiDocs
-
-      - name: Build OpenAPI Spec Protected
-        working-directory: lapis
-        run: ./gradlew generateOpenApiDocs -PopennessLevel=protected
-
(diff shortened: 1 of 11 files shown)


fengelniederhammer:
> LGTM. It works for me. However, I am not sure if this is linked to this issue or not: When I ue the http://localhost:8090/swagger-ui/index.html#/single-segmented-sequence-controller/getAlignedNucleotideSequences I have to select something at `orderBy`, otherwise I don't get any data. Is this related to swagger or a feature?

I checked the logs: Selecting not filters and leaving all other values at their defaults, it sends `{"action":{"randomize":false,"limit":100,"type":"FastaAligned","sequenceName":"main"},"filterExpression":{"type":"True"}}` to SILO, which seems correct to me. -> not a LAPIS issue, but something in SILO (Probably https://github.com/GenSpectrum/LAPIS-SILO/issues/785)

PR #1246, opened 2025-06-24
feat(lapis): add endpoint `/sample/unalignedNucleotideSequences` for multi-segmented organism to download several sequences at once

### LINKED ISSUE #1223: Add endpoints for getting a FASTA with multiple segments
* `/sample/unalignedNucleotideSequences?segment=S&segment=F`
* the same for aligned sequences and amino acid sequences
* If no segment/gene specified, one get all segments
* FASTA header: `><primary key>|<segment name>`

Q:
* This clashes a bit with the current `/sample/alignedNucleotideSequences` endpoints. How do we avoid that? 
* Should it also be enabled for single segmented organisms?

### LINKED ISSUE #1236: Multi segmented fasta endpoint: unaligned nucleotide sequences
Part 2: unaligned nucleotide sequences

fengelniederhammer:
resolves #1236
resolves #1223

resolves #

## PR Checklist

### DIFF
--- a/lapis-e2e/test/details.spec.ts
+++ b/lapis-e2e/test/details.spec.ts
@@ -253,7 +253,7 @@ key_1002052
   it('should order by random', async () => {
     const result = await lapisClient.postDetails({
       detailsPostRequest: {
-        orderBy: [{ field: 'random' }, { field: 'division', type: 'descending' }],
+        orderBy: [{ field: 'random' }, { field: 'division', type: 'ascending' }],
         fields: ['primaryKey', 'division'],
       },
     });


lapis/src/main/kotlin/org/genspectrum/lapis/config/ReferenceGenome.kt
```diff
@@ -36,6 +36,11 @@ class ReferenceGenomeSchema(
 
     fun isSingleSegmented(): Boolean = nucleotideSequences.size == 1
 
+    fun getSequenceNameFromCaseInsensitiveName(name: String) =
+        nucleotideSequenceNames[name.lowercase()]?.name
+            ?: geneNames[name.lowercase()]?.name
+            ?: name
```
chaoran-chen, line 42: Is `?: name` an expected case or rather an error?

fengelniederhammer: The code also pipes metadata field names through here. Maybe this is a little hacky, but I still think that a reasonable default makes more sense than an error here.

chaoran-chen (approved):
Thanks!

PR #1258, opened 2025-07-04
feat(lapis): add parameter `fastaHeaderTemplate` to customize the fasta header in sequences endpoints

### LINKED ISSUE #857: FASTA descriptions
(Assuming that it doesn't) it would be nice if LAPIS could support returning FASTAs with "descriptions", e.g. https://www.ncbi.nlm.nih.gov/nuccore/NC_045512.2?report=fasta has both an accession and then a description, separated by a space.

https://github.com/loculus-project/loculus/issues/2284#issuecomment-2220819967

fengelniederhammer:
resolves #857

This got a bit larger than expected. Essentially how it works:
* parse the template, map the fields to the names that they have in the config (to make them case insensitive) and also validate them in the process
* pass the detected fields to SILO in the `additionalFields` field (still requires https://github.com/GenSpectrum/LAPIS-SILO/issues/860 so that it also works for unaligned sequences)
* change the sequences response that we expect in LAPIS to a generic map
* hand this map to the template to fill the placeholders

When requesting a data format that is not Fasta, then it will be ignored. The previous header format is now simply the default value.

## PR Checklist

### DIFF
--- a/lapis/src/test/kotlin/org/genspectrum/lapis/controller/MultiSegmentedSequenceControllerTest.kt
+++ b/lapis/src/test/kotlin/org/genspectrum/lapis/controller/MultiSegmentedSequenceControllerTest.kt
@@ -279,7 +279,7 @@ class MultiSegmentedSequenceControllerTest(
         every {
             siloQueryModelMock.getGenomicSequence(
                 sequenceFilters = sequenceFiltersRequestWithSegments(
-                    sequenceFilters = mapOf("country" to "Germany"),
+                    sequenceFilters = mapOf("country" to "Switzerland"),
                     segments = listOf(SEGMENT_NAME, otherSegment),
                 ),
                 sequenceType = SequenceType.UNALIGNED,


lapis/src/main/kotlin/org/genspectrum/lapis/controller/ControllerDescriptions.kt
```diff
@@ -79,6 +79,18 @@ Alternatively, the data format can be specified by setting the \"Accept\"-header
 When both are specified, the request parameter takes precedence over the header.
 """
 
+const val NUCLEOTIDE_FASTA_HEADER_TEMPLATE_DESCRIPTION = """Use this to customize the header of the FASTA file.
+This parameter will be ignored if the data format is not FASTA.
+It supports free-form text, placeholders of the form "{metadataField}" to include metadata fields in the header and
+"{.segment}" to include the segment name in the header.
```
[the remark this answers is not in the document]
chaoran-chen: This description is actually inaccurate, people should write `.segment` as a keyword if they want to have the segment name. 

```suggestion
".segment" to include the segment name in the header.
```

The dot is intentional to avoid introducing even more names that may not be used for metadata fields because they'd conflict with reserved keywords.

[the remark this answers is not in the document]
anna-parker: I would explicitly say `add the keyword "{.segment}" to include..`

lapis/src/main/kotlin/org/genspectrum/lapis/response/SequencesStreamer.kt
```diff
@@ -1,79 +1,65 @@
...
 import org.springframework.stereotype.Component
 import java.nio.charset.Charset
-import java.util.stream.Stream
 
 @Component
 class SequencesStreamer(
     private val dataVersion: DataVersion,
     private val objectMapper: ObjectMapper,
 ) {
     fun stream(
```
anna-parker, line 20: The refactor of `SequencesStreamer.stream` seems unrelated to this feature - it might be good to split it out into a separate PR to make review easier

lapis/src/main/kotlin/org/genspectrum/lapis/openApi/Schemas.kt
```diff
@@ -469,7 +473,23 @@ annotation class DataFormat
     schema = Schema(ref = "#/components/schemas/$SEQUENCES_FORMAT_SCHEMA"),
     description = SEQUENCES_DATA_FORMAT_DESCRIPTION,
 )
-annotation class SequencesDataFormat
+annotation class SequencesDataFormatParam
```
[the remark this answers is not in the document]
fengelniederhammer: There is a name clash otherwise and I figured that I would rather rename this instead of the other one.

lapis/src/main/kotlin/org/genspectrum/lapis/controller/LapisController.kt
```diff
@@ -1705,17 +1711,24 @@ class LapisController(
...
 
+        val sequencesDataFormat = SequencesDataFormat.fromAcceptHeaders(httpHeaders.accept)
+
         siloQueryModel.getGenomicSequence(
             sequenceFilters = request,
             sequenceType = SequenceType.ALIGNED,
             sequenceNames = genes ?: referenceGenomeSchema.getGeneNames(),
+            rawFastaHeaderTemplate = getFastaHeaderTemplate(
+                requestedTemplate = fastaHeaderTemplate,
+                defaultTemplate = "{${databaseConfig.schema.primaryKey}}|{.gene}",
```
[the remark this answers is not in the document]
fengelniederhammer: No, there should not be a change in the default behavior (if there is, then it is a mistake and needs to be fixed in this PR ;) )

lapis-e2e/test/alignedNucleotideSequence.spec.ts
```diff
@@ -143,5 +169,64 @@ describe('The /alignedNucleotideSequence endpoint', () => {
...
+      expect(result).to.have.length(6);
+      expect(
+        Object.entries(result[0])
+          .filter(([_, value]) => value !== undefined)
+          .map(([key]) => key)
+      ).to.have.members(['primaryKey', 'm']);
+    });
+
+    it('should ignore the fasta header template when returning all sequences JSON', async () => {
+      const result = await lapisMultiSegmentedSequenceController.postAllAlignedNucleotideSequences({
```
anna-parker, line 191: shouldn't this raise an error if the `fastaHeaderTemplate` contains invalid metadata fields?

lapis/src/main/kotlin/org/genspectrum/lapis/controller/middleware/DataFormatParameterFilter.kt
```diff
@@ -23,10 +24,32 @@ object DataFormat {
     const val TSV = "TSV"
 }
 
-object SequencesDataFormat {
-    const val JSON = DataFormat.JSON
-    const val FASTA = "FASTA"
-    const val NDJSON = "NDJSON"
+enum class SequencesDataFormat(
```
anna-parker, line 27: Again I think it would be good to move this refactor to a separate PR

lapis-e2e/test/alignedNucleotideSequence.spec.ts
fengelniederhammer: The purpose of the test is to show that the template is ignored when you do not request fasta format.

lapis-e2e/test/alignedNucleotideSequence.spec.ts
chaoran-chen, line 1: Thanks very much! I tested the instance on staging and it worked very well! Just one thing: if a value is `null`, then it currently also writes `null`, for example, I got `>MT641718|AUS/NT27/2020|null|Australia` because `date` was null. I think that it would be better to write an empty string instead, i.e., output ``>MT641718|AUS/NT27/2020||Australia`.

lapis/src/main/kotlin/org/genspectrum/lapis/model/SiloQueryModel.kt
```diff
@@ -193,24 +195,41 @@ class SiloQueryModel(
...
-                limit = sequenceFilters.limit,
-                offset = sequenceFilters.offset,
+        rawFastaHeaderTemplate: String,
+        sequenceSymbolType: SequenceSymbolType,
+    ): SequencesResponse {
+        val fastaHeaderTemplate = fastaHeaderTemplateParser.parseTemplate(
+            template = rawFastaHeaderTemplate,
+            sequenceSymbolType = sequenceSymbolType,
+        )
+        val cleanedSequenceNames = sequenceNames
```
anna-parker, line 205: It looks like this isn't handled at the moment either but if `sequenceNames` contains a value that is not in the `referenceGenomeSchema` it is just returned without an error - shouldn't we raise an error at this point? At the moment we just pass this on to SILO

lapis/src/main/kotlin/org/genspectrum/lapis/response/SiloResponse.kt
```diff
@@ -97,47 +96,10 @@ class SequenceDataDeserializer(
...
 
         return SequenceData(
-            sequenceKey = sequenceKey,
-            sequences = sequences,
+            node.properties().associate { (key, value) -> key.removePrefix(UNALIGNED_PREFIX) to value },
         )
     }
 }
-
-@JsonComponent
```
anna-parker, line 123: was this function not used before?

fengelniederhammer: > So this PR is changing how we do serialization. Do we have tests for our serialization? I haven't looked into how this is done for LAPIS - in SILO we have to update the serialization version.

The serialization version in SILO is for the binary output of the preprocessing. That is unrelated to JSON serialization.

We have a lot of tests that check the response of LAPIS (unit and e2e). All of them implicitly test the serialization.
 
> Again this is a refactor that should be made in a separate PR to make review easier.

IMO that is not a refactoring. There is a change in requirements that this PR adapts the code to (previously: every set of sequences is the primary key along with a number of `(sequenceName, sequence)` pairs - now: we only have a generic map, because we know neither the metadata nor the sequences that are contained in the response).
Doing this in a separate PR would require quite a number of adaptions, too. Most of them would be obsolete again with this PR. I don't think it's worth the additional effort of artificially splitting it.

lapis-e2e/test/alignedNucleotideSequence.spec.ts
```diff
@@ -143,5 +169,64 @@ describe('The /alignedNucleotideSequence endpoint', () => {
...
+        primaryKey: 'key_0',
+      });
+
+      const response = await fetch(
+        `${basePathMultiSegmented}/sample/alignedNucleotideSequences/M?${urlParams}`
+      );
+
+      const text = await response.text();
+      expect(response.status, text).to.equal(200);
+      expect(text.split('\n')[0]).to.equal('>key=key_0|2021-03-18|Switzerland|M');
```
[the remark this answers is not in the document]
fengelniederhammer: It's probably not obvious, but there are unit tests that cover the case: https://github.com/GenSpectrum/LAPIS/blob/36754ac59d6b5135a86094861994a952575f7edf/lapis/src/test/kotlin/org/genspectrum/lapis/controller/MockData.kt#L209-L235

Is that enough?

lapis-e2e/test/alignedNucleotideSequence.spec.ts
fengelniederhammer: I changed it. Now it throws an error.

lapis-e2e/test/alignedNucleotideSequence.spec.ts
```diff
@@ -170,33 +171,34 @@ describe('The /alignedNucleotideSequence endpoint', () => {
...
     });
 
-    it('should ignore the fasta header template when returning JSON', async () => {
-      const result = await lapisMultiSegmentedSequenceController.postAlignedNucleotideSequence({
-        nucleotideSequenceRequest: {
-          dataFormat: 'JSON',
-          fastaHeaderTemplate: '{primaryKey}{date}{something invalid}',
-        },
-        segment: 'M',
+    it('should throw an error for fasta header template when returning JSON', async () => {
```
anna-parker, line 174: I guess this function is now a duplication of the one above so we can remove it :-) (surprised there isn't a warning that two functions have the same name)

anna-parker: Ah sorry I just saw this is for the multi-segmented case so you can ignore me

lapis/src/test/kotlin/org/genspectrum/lapis/silo/SiloClientTest.kt
```diff
@@ -169,9 +170,9 @@ class SiloClientTest(
         assertThat(
             result,
             containsInAnyOrder(
-                SequenceData(sequenceKey = "key1", sequences = mapOf("someSequenceName" to "ABCD")),
-                SequenceData(sequenceKey = "key2", sequences = mapOf("someSequenceName" to "DEFG")),
-                SequenceData(sequenceKey = "key3", sequences = mapOf("someSequenceName" to null)),
+                SequenceData(mapOf("primaryKey" to TextNode("key1"), "someSequenceName" to TextNode("ABCD"))),
```
anna-parker, line 173: hmmm... can SILO ever return multiple segment sequences here? Maybe we should add a test for that

lapis/src/test/kotlin/org/genspectrum/lapis/model/SiloQueryModelTest.kt
```diff
@@ -64,10 +70,36 @@ class SiloQueryModelTest {
 
     private lateinit var underTest: SiloQueryModel
 
+    private val fastaHeaderTemplateParser = FastaHeaderTemplateParser(
+        caseInsensitiveFieldsCleaner = CaseInsensitiveFieldsCleaner(
+            databaseConfig = DatabaseConfig(
```
anna-parker, line 75: is there a reason why you didn't use the `databaseConfigWithFields`?

lapis/src/test/kotlin/org/genspectrum/lapis/silo/SiloClientTest.kt
anna-parker: Ah I see you did this in `lapis/src/test/kotlin/org/genspectrum/lapis/controller/MockData.kt` so feel free to ignore me :-)

lapis/src/test/kotlin/org/genspectrum/lapis/model/FastaHeaderTemplateTest.kt
```diff
@@ -0,0 +1,257 @@
...
+import org.genspectrum.lapis.controller.BadRequestException
+import org.genspectrum.lapis.request.CaseInsensitiveFieldsCleaner
+import org.hamcrest.CoreMatchers.`is`
+import org.hamcrest.MatcherAssert.assertThat
+import org.hamcrest.Matchers.containsString
+import org.junit.jupiter.api.Test
+import org.junit.jupiter.api.assertThrows
+
+class FastaHeaderTemplateTest {
+    private val fastaHeaderTemplateParser = FastaHeaderTemplateParser(
```
anna-parker, line 22: This looks like a copy of the template parser in `lapis/src/test/kotlin/org/genspectrum/lapis/model/SiloQueryModelTest.kt` can we reuse?

lapis/src/test/kotlin/org/genspectrum/lapis/controller/MultiSegmentedSequenceControllerTest.kt
```diff
@@ -257,8 +284,14 @@ class MultiSegmentedSequenceControllerTest(
...
                 sequenceType = SequenceType.UNALIGNED,
                 sequenceNames = listOf(SEGMENT_NAME, otherSegment),
+                rawFastaHeaderTemplate = "{primaryKey}|{.segment}",
+                sequenceSymbolType = SequenceSymbolType.NUCLEOTIDE,
             )
-        } returns Stream.empty()
+        } returns SequencesResponse(
+            sequenceData = Stream.empty(),
+            requestedSequenceNames = listOf(SEGMENT_NAME, otherSegment),
+            fastaHeaderTemplate = FastaHeaderTemplate("", emptySet()),
```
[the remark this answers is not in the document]
fengelniederhammer: It's a mock response. There is no technical requirement that it follows any logic. `FastaHeaderTemplateParser.parseTemplate` is not called in this test (only the mock is called). I used this because it was the easiest to implement.
But I can change it if it confuses you too much.

lapis/src/test/kotlin/org/genspectrum/lapis/model/FastaHeaderTemplateTest.kt
fengelniederhammer: Hm, we could. But I think I'd rather leave them separate.

One reason why you usually try to reuse stuff is so that you don't have to do the same changes in multiple places. but those are separate tests, they can have different requirements.

But I think it makes sense to a least extract something for the verbose database config, I'll do that.

PR #1277, opened 2025-07-22
feat(lapis): add mostRecentCommonAncestor endpoint

anna-parker:
resolves https://github.com/GenSpectrum/LAPIS/issues/1279

Adds a LAPIS endpoint for the new SILO endpoint getMostRecentCommonAncestor defined here: https://github.com/GenSpectrum/LAPIS-SILO/blob/main/documentation/phylogenetic_queries.md

The endpoint expects a query with the structure: 
```
"action": {
  "type": "MostRecentCommonAncestor",
  "columnName": "COLUMN_NAME",
  "printNodesNotInTree": true
}
```
where columnName must be a column with the `phyloTreeNodeIdentifier=true` property in the config. `columnName` is a required field in the LAPIS endpoint I call this endpoint `phyloTreeField` to be clearer (maybe we should also change this in SILO: https://github.com/GenSpectrum/LAPIS-SILO/issues/906), the optional `printNodesNotInTree` field is used as is. 

As `phyloTreeField` is a required field it had to be added to all test endpoint calls which lead to there being quite a few changes in the code. However, now it should be covered by all matrix tests. Although we did not initially plan for the endpoint to also allow tsv and csv response type I added these options as it was easier than refactoring the matrix tests. I also added a few additional tests to check that the error messages when `phyloTreeField` is not set correctly are clear.

For the e2e tests I generated a nwk tree from the example data and added docs on how this can be replicated.

## PR Checklist
- [ ] Follow up PR: document how to add a new endpoint to LAPIS


lapis/src/main/kotlin/org/genspectrum/lapis/openApi/OpenApiDocs.kt
```diff
@@ -279,6 +293,17 @@ fun buildOpenApiSchema(
...
                 .addSchemas(DETAILS_FIELDS_SCHEMA, fieldsArray(databaseConfig.schema.metadata))
+                .addSchemas(
+                    PRINT_NODES_NOT_IN_TREE_FIELD_SCHEMA,
+                    Schema<Boolean>(),
+                )
+                .addSchemas(
+                    PHYLO_TREE_FIELD_SCHEMA,
+                    fieldsArray(
+                        databaseConfig.schema.metadata
+                            .filter { it.phyloTreeNodeIdentifier },
```
anna-parker, line 304: there are no options shown in the swagger now...

lapis-e2e/test/mostRecentCommonAncestor.spec.ts
```diff
@@ -0,0 +1,89 @@
+import { expect } from 'chai';
+import { basePath, lapisClient } from './common';
+
+describe('The /mostRecentCommonAncestor endpoint', () => {
+  it('should return MRCA with missing nodes if specified', async () => {
+    const result = await lapisClient.postMostRecentCommonAncestor({
+      mostRecentCommonAncestorResponse: {
```
anna-parker, line 7: the contents do not appear to be passed to the endpoint... its also weird that the openAPI generated library calls this field response when it should be for the request...

fengelniederhammer: The naming issue is probably one that you introduced somewhere in LAPIS. The generator in TS doesn't make up names, it takes them from the spec.

Did you debug the TS code here while running the test? That's not difficult and normally already points you to the problem rather quickly.

lapis/src/main/kotlin/org/genspectrum/lapis/controller/LapisController.kt
```diff
@@ -894,6 +899,253 @@ class LapisController(
...
+            response = response,
+            responseFormat = ResponseFormat.Json,
+        )
+    }
+
+    @GetMapping(MOST_RECENT_COMMON_ANCESTOR_ROUTE, produces = [TEXT_CSV_VALUE])
+    @StringResponseOperation(
+        operationId = "getMostRecentCommonAncestorAsCsv",
+    )
+    fun getMostRecentCommonAncestorAsCsv(
```
fengelniederhammer, line 1033: Nitpick: Could you please reorder the methods? We usually have the GET methods before the POST methods

lapis/src/main/kotlin/org/genspectrum/lapis/request/Field.kt
```diff
@@ -9,6 +9,10 @@ data class Field(
 
 fun interface FieldConverter<T> {
     fun convert(source: String): T
+
+    fun validatePhyloTreeFields(source: String) {
+        // Default implementation does nothing, can be overridden if needed
+    }
```
fengelniederhammer, line 15: Do we really need this method in the interface here? To me it looks like we don't

lapis-e2e/test/mostRecentCommonAncestor.spec.ts
fengelniederhammer: image
That also doesn't look right ;)

lapis/src/main/kotlin/org/genspectrum/lapis/request/Field.kt
```diff
@@ -23,4 +27,17 @@ class CaseInsensitiveFieldConverter(
...
+        val validFields = caseInsensitiveFieldsCleaner.getPhyloTreeFields()
+        if (converted.fieldName !in validFields) {
+            throw BadRequestException(
+                "Field '${converted.fieldName}' is not a phylo tree field, " +
+                    "known phylo tree fields are [${validFields.joinToString(
+                        ", ",
+                    )}]",
+            )
+        }
+    }
```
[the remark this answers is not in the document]
anna-parker: It is to validate that the user passes a field for phyloTreeField that can be used for phylo tree queries - when calling the endpoint outside swagger it is possible to enter any valid metadata field but this will mean that SILO will return an error

lapis-e2e/test/mostRecentCommonAncestor.spec.ts
anna-parker: Thanks this helped me find the bug - I think it is now fixed :-)

lapis/src/main/kotlin/org/genspectrum/lapis/request/SequenceFiltersRequestWithFields.kt
```diff
@@ -48,6 +61,36 @@ class SequenceFiltersRequestWithFieldsDeserializer(
     }
 }
 
+@JsonComponent
+class PhyloTreeSequenceFiltersRequestWithFieldsDeserializer(
+    private val caseInsensitiveFieldConverter: CaseInsensitiveFieldConverter,
```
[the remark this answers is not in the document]
anna-parker: Oh odd if I remove both I get a build error here: file:///Users/aparker/Documents/Github/LAPIS/lapis/src/main/kotlin/org/genspectrum/lapis/request/MutationProportionsRequest.kt:67:22 Unresolved reference 'parseFieldsProperty'.
e: file:///Users/aparker/Documents/Github/LAPIS/lapis/src/main/kotlin/org/genspectrum/lapis/request/MutationProportionsRequest.kt:67:76 Unresolved reference 'it'.

Just removing the `PhyloTreeSequenceFiltersRequestDeserializer` doesn't cause a build error but triggers 46 errors in the tests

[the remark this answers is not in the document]
anna-parker: e.g. 
```
LapisControllerDownloadAsFileTest > WHEN I request compressed files THEN the filenames have a corresponding suffix(DownloadCompressedFileScenario) > POST form url encoded /mostRecentCommonAncestor as zstd tsv should return compressed file FAILED
    java.lang.AssertionError: Status expected:<200> but was:<400>
        at org.springframework.test.util.AssertionErrors.fail(AssertionErrors.java:61)
        at org.springframework.test.util.AssertionErrors.assertEquals(AssertionErrors.java:128)
        at org.springframework.test.web.servlet.result.StatusResultMatchers.lambda$matcher$9(StatusResultMatchers.java:640)
        at org.springframework.test.web.servlet.MockMvc$1.andExpect(MockMvc.java:214)
        at org.genspectrum.lapis.controller.LapisControllerDownloadAsFileTest.WHEN I request compressed files THEN the filenames have a corresponding suffix(LapisControllerDownloadAsFileTest.kt:154)
        ```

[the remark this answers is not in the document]
Comment (anna-parker) [reply lapis/src/main/kotlin/org/genspectrum/lapis/request/SequenceFiltersRequestWithFields.kt]: Closer look shows " Body = {"type":"about:blank","title":"Bad Request","status":400,"detail":"Failed to read request","instance":"/sample/mostRecentCommonAncestor"}" - so it does appear to be required for reading the SILO response

File: lapis/src/main/kotlin/org/genspectrum/lapis/request/Field.kt (hunk 5 above)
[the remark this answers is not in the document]
Comment (anna-parker) [reply lapis/src/main/kotlin/org/genspectrum/lapis/request/Field.kt:40]: line 32

File: lapis/src/main/kotlin/org/genspectrum/lapis/controller/LapisController.kt [hunk 7]
```diff
@@ -894,6 +899,255 @@ class LapisController(
...
```
fengelniederhammer, line 956: I mean to recall that we said that we don't need CSV for this action since it always only returns a single line? Isn't that true anymore? 

If it was only ever a single line, then it would also make more sense to return no array, but an object.

I.e.:
```
{
  "data": {
      "mrcaNode": "NODE_0000000",
      "missingNodeCount": 2,
      "missingFromTree": null
  }
  ,
  "info": {...}
}
```

instead of 
```
{
  "data": [
    {
      "mrcaNode": "NODE_0000000",
      "missingNodeCount": 2,
      "missingFromTree": null
    }
  ],
  "info": {...}
}
```

fengelniederhammer (changes requested):
I think when the filter matches no sequence, then it returns a 500 error:
```
{
  "primaryKey": "string",
"phyloTreeField":   "primaryKey"
}
```
yields
```
{
  "error": {
    "type": "about:blank",
    "title": "Internal Server Error",
    "status": 500,
    "detail": "Could not parse response from silo: class com.fasterxml.jackson.module.kotlin.MissingKotlinParameterException Instantiation of [simple type, class org.genspectrum.lapis.response.MostCommonAncestorData] value failed for JSON property mrcaNode due to missing (therefore NULL) value for creator parameter mrcaNode which is a non-nullable type\n at [Source: REDACTED (`StreamReadFeature.INCLUDE_SOURCE_IN_LOCATION` disabled); line: 1, column: 38] (through reference chain: org.genspectrum.lapis.response.MostCommonAncestorData[\"mrcaNode\"]) - NDJSON line: {\"mrcaNode\":null,\"missingNodeCount\":0}"
  },
  "info": {
    "dataVersion": null,
    "requestId": "a7d3b491-ce5f-427c-9e29-acc843970171",
    "requestInfo": "sars_cov-2_minimal_test_config on localhost at 2025-07-24T09:13:38.580825235",
    "reportTo": "Please report to https://github.com/GenSpectrum/LAPIS/issues in case you encounter any unexpected issues. Please include the request ID and the requestInfo in your report.",
    "lapisVersion": "local",
    "siloVersion": "0.7.5"
  }
}
```

lapis-e2e/testData/multiSegmented/testDatabaseConfig.yaml
```diff
@@ -4,6 +4,7 @@ schema:
   metadata:
     - name: primaryKey
       type: string
+      phyloTreeNodeIdentifier: true
```
fengelniederhammer, line 7: Do we actually have a tree for this case? 

What happens in that case actually? If you don't have a tree in the data, but you still call that endpoint? Do you get an empty response?

lapis/src/main/kotlin/org/genspectrum/lapis/openApi/OpenApiDocs.kt
```diff
@@ -423,6 +462,16 @@ private fun detailsMetadataFieldSchemas(databaseConfig: DatabaseConfig) =
...
         .associate { it.name to Schema<String>().types(setOf(mapToOpenApiType(it.type))) }
 
+private fun mostRecentCommonAncestorMetadataFieldSchemas(): Map<String, Schema<*>> =
+    mapOf(
+        "mrcaNode" to StringSchema()
+            .description("The label of the most recent common ancestor node in the phylogenetic tree."),
+        "missingNodeCount" to IntegerSchema()
+            .description("The number of nodes in the filter not in the phylogenetic tree."),
+        "missingFromTree" to StringSchema()
+            .description("The comma-separated list of names of the nodes in the filter not in the phylogenetic tree."),
```
fengelniederhammer, line 472: If we already know that this is a list - why don't we return a list then? Is there a good reason to return a csv string instead?

lapis/src/main/kotlin/org/genspectrum/lapis/openApi/OpenApiDocs.kt
```diff
@@ -196,6 +211,17 @@ fun buildOpenApiSchema(
...
                     ),
                 )
+                .addSchemas(
+                    MOST_RECENT_COMMON_ANCESTOR_RESPONSE_SCHEMA,
+                    lapisArrayResponseSchema(
+                        Schema<String>()
+                            .types(setOf("object"))
+                            .description(
+                                "The response contains the most recent common ancestor node label `mrcaNode`, the number of nodes in the filter not in the phylotree `missingNodeCount` and optionally their names `missingFromTree`.",
+                            )
```
fengelniederhammer, line 221: Do we need this description? AFAICS all of the information here is also already in the description of the fields of this object.

lapis/src/main/kotlin/org/genspectrum/lapis/request/CaseInsensitiveFieldsCleaner.kt
```diff
@@ -9,7 +9,11 @@ class CaseInsensitiveFieldsCleaner(
 ) {
     private val fieldsMap = databaseConfig.schema.metadata.map { it.name }.associateBy { it.lowercase() }
 
+    private val phyloTreeFields = databaseConfig.schema.metadata.filter { it.phyloTreeNodeIdentifier }.map { it.name }
```
fengelniederhammer, line 12: Nitpick: I feel that this field is somewhat misplaced in this class. Why is it here? What does it have to do with the things that are already in this class? It could as well be in the place that uses `getPhyloTreeFields`. (It also doesn't necessarily need a getter for a simple field that could be public)

lapis/src/main/kotlin/org/genspectrum/lapis/request/Field.kt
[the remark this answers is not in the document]
fengelniederhammer: But the calling side doesn't get the converted fields. We need to send the converted fields to SILO.

I think it's worth to add an e2e test that checks that case (if there isn't already one, I haven't looked at them yet)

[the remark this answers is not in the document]
anna-parker: I also saw this and updated the code and added a unit test :-)

lapis/src/main/kotlin/org/genspectrum/lapis/openApi/OpenApiDocs.kt
fengelniederhammer: Why does SILO return a string then when it already knows that it's a list?

lapis/src/main/kotlin/org/genspectrum/lapis/request/CaseInsensitiveFieldsCleaner.kt
anna-parker: This seemed the neatest way to do this - otherwise we have to pass the databaseConfig everywhere

lapis/src/main/kotlin/org/genspectrum/lapis/controller/LapisController.kt
anna-parker: This might be nicer but it would require a new LapisResponseStreamer which doesn't expect a list of records - so I would prefer to refactor this in a later PR as this is already getting really long. I'm also not sure if it might be better to have the same output structure for all requests - @chaoran-chen what do you think the user will expect?

lapis-e2e/testData/multiSegmented/testDatabaseConfig.yaml
anna-parker: SILO throws an error: "MRCA action cannot be called on Column '{}' as it does not have a phylogenetic tree associated with it" - as the unit tests actually never call SILO we don't need to add a tree for the unit tests - I added one for the e2e tests :-)

lapis/src/main/kotlin/org/genspectrum/lapis/controller/LapisController.kt
chaoran-chen: Both solutions are fine and I'm happy with keeping `data: []` since it is already implemented. Taking the first element is an acceptable burden and if because `data: []` is conform with the other endpoints, people might be able to reuse code more easily.

lapis/src/main/kotlin/org/genspectrum/lapis/openApi/OpenApiDocs.kt
anna-parker: @Taepper ah so SILO doesn't currently support nested types? 

Then lets keep it like this for now :-)

fengelniederhammer: Do you think it makes sense to do the mapping in LAPIS then for now (csv string -> list)? (Not necessarily in this PR though)

lapis/src/main/kotlin/org/genspectrum/lapis/controller/LapisController.kt
fengelniederhammer: Then let's create an issue and discuss in the next Tuesday meeting whether we want to do it?

lapis/src/main/kotlin/org/genspectrum/lapis/openApi/OpenApiDocs.kt
anna-parker: So tbh I think its quite unlikely that there will be any nodes in this field (unless our tree is very out of sync with our data - and I think we should anyways try to avoid that) so I'm not sure it makes much sense to format it much

fengelniederhammer: Will it be used? If yes, then I think we should provide a proper API, especially since it seems very easy in this case. If no, why do we have it at all?

fengelniederhammer: Then let's create an issue so that we don't forget about it?

lapis/src/main/kotlin/org/genspectrum/lapis/request/CaseInsensitiveFieldsCleaner.kt
fengelniederhammer: IMO "passing the database config everywhere" doesn't matter. It's Spring bean, it's very cheap to distribute it.

I would rather keep the responsibilities of classes separated properly. We can simply move this method over to the `CaseInsensitiveFieldConverter` and it's done.

lapis/src/main/kotlin/org/genspectrum/lapis/openApi/OpenApiDocs.kt
anna-parker: https://github.com/GenSpectrum/LAPIS/issues/1285

lapis/src/main/kotlin/org/genspectrum/lapis/request/PhyloTreeSequenceFiltersRequest.kt
```diff
@@ -0,0 +1,60 @@
...
+
+fun <T> parsePhyloTreeProperty(
+    node: JsonNode,
+    fieldConverter: FieldConverter<T>,
+): T {
+    val phyloTreeField = node.get(PHYLO_TREE_FIELD_PROPERTY)
+    if (phyloTreeField == null) {
+        "$PHYLO_TREE_FIELD_PROPERTY is required and must be a string representing a phylo tree field"
+    }
+    return fieldConverter.validatePhyloTreeFields(phyloTreeField.asText())
```
fengelniederhammer, line 59: We should also check that it's a string node here.
```suggestion
    val phyloTreeField = node.get(PHYLO_TREE_FIELD_PROPERTY)
    if (phyloTreeField == null) {
        "$PHYLO_TREE_FIELD_PROPERTY is required and must be a string representing a phylo tree field"
    }
    if (!phyloTreeField.isTextual) {
        throw BadRequestException(
            "$PHYLO_TREE_FIELD_PROPERTY must be a string, but was ${phyloTreeField.nodeType}",
        )
    }
    return fieldConverter.validatePhyloTreeFields(phyloTreeField.textValue())
```

fengelniederhammer (changes requested):
Some details, otherwise it looks good to me.

lapis/src/test/kotlin/org/genspectrum/lapis/silo/SiloQueryTest.kt
```diff
@@ -40,6 +40,31 @@ class SiloQueryTest {
         assertThat(objectMapper.readTree(result), equalTo(objectMapper.readTree(expected)))
     }
 
+    @Test
+    fun `MRCA query is correctly serialized to JSON`() {
```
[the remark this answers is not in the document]
anna-parker: Ah I just added a test similar to the `Query is correctly serialized to JSON` test above but yes I can move this to the `getTestSiloActions`.

lapis/src/main/kotlin/org/genspectrum/lapis/request/Field.kt
```diff
@@ -9,6 +9,8 @@ data class Field(
 
 fun interface FieldConverter<T> {
     fun convert(source: String): T
+
+    fun validatePhyloTreeFields(source: String): T = convert(source)
```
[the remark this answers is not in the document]
anna-parker: I just tried out the second option - it doesn't work too well as `validatePhyloTreeField` is also used in the LapisController before the Deserializer is used

lapis/src/test/kotlin/org/genspectrum/lapis/controller/LapisControllerPhyloTreeTest.kt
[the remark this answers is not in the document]
fengelniederhammer: As long as the errors are clear to the user, that's ok.

lapis/src/main/kotlin/org/genspectrum/lapis/request/Field.kt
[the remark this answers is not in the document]
anna-parker: Ok! That works! Let me know what you think of the changes in https://github.com/GenSpectrum/LAPIS/pull/1277/commits/1573cb47048db6f03a7a00e9bfea1eaeff157e77 :-)

fengelniederhammer (approved):
LGTM :)

PR #1281, opened 2025-07-24
refactor(lapis): also reject unknown properties when reading the other reference genomes

fengelniederhammer:
## PR Checklist


fengelniederhammer:
> Why would we like to do that? I usually prefer being strict when it comes to config files / developer-facing input.

We partially already read that file with ignoring additional properties, but not fully.

I got an error in Loculus because the reference genomes for LAPIS still had the INSDC accession field that we have there (because I forgot to clean that before giving it to LAPIS on my branch). I found it more annoying than helpful in this case.

In general I agree that it makes more sense to be strict. But in this environment we have a lot of config files that are distributed across different artifacts. If one artifact needs a special additional property (such as the INSDC accession in Loculus), then it makes sense to be more relaxed. We already do that for the database config anyway because SILO has some fields that LAPIS doesn't have and vice versa.

chaoran-chen (approved):

PR #1288, opened 2025-07-29
refactor(lapis): fix condition on cacheable on nullable boolean

### LINKED ISSUE #1278: SiloAction with `override val randomize: Boolean? = null` raises spring serialization error
Although we set `override val randomize: Boolean? = null` as a default in all action constructors actually using the `null` default raises an error:
```
.ExceptionHandler: Caught unexpected exception: EL1001E: Type conversion problem, cannot convert from null to boolean
org.springframework.expression.spel.SpelEvaluationException: EL1001E: Type conversion problem, cannot convert from null to boolean
```
It seems although this is set as a default all functions actually pass a boolean for randomize so we weren't encountering this issue until I added a new endpoint

fengelniederhammer:
resolves #1278

The problem was that Spring expression, not JSON serialization.

## PR Checklist

### DIFF
--- a/lapis/src/main/kotlin/org/genspectrum/lapis/silo/SiloClient.kt
+++ b/lapis/src/main/kotlin/org/genspectrum/lapis/silo/SiloClient.kt
@@ -80,7 +80,7 @@ open class CachedSiloClient(
 ) {
     private val httpClient = HttpClient.newHttpClient()
 
-    @Cacheable(SILO_QUERY_CACHE_NAME, condition = "#query.action.cacheable && !#query.action.randomize")
+    @Cacheable(SILO_QUERY_CACHE_NAME, condition = "#query.action.cacheable && !(#query.action.randomize ?: false)")
     open fun <ResponseType> sendCachedQuery(query: SiloQuery<ResponseType>): WithDataVersion<List<ResponseType>> =
         sendQuery(query)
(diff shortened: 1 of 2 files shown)


lapis/src/main/kotlin/org/genspectrum/lapis/silo/SiloClient.kt
```diff
@@ -80,7 +80,7 @@ open class CachedSiloClient(
 ) {
     private val httpClient = HttpClient.newHttpClient()
 
-    @Cacheable(SILO_QUERY_CACHE_NAME, condition = "#query.action.cacheable && !#query.action.randomize")
+    @Cacheable(SILO_QUERY_CACHE_NAME, condition = "#query.action.cacheable && !(#query.action.randomize ?: false)")
```
anna-parker, line 83: Just to check that I understand the effect of the `randomize` parameter - this actually affects if the query is cached? Maybe we should add an explanation somewhere as I don't find this name very intuitive I thought it was a way to specify if the output order was random

anna-parker: Maybe renaming it to `disableCache` would actually be better

anna-parker: Ah ok! Yes that makes sense thanks for the clarification!

PR #1290, opened 2025-07-29
chore(lapis): fix compiler warnings

fengelniederhammer:
It said:
"This annotation is currently applied to the value parameter only, but in the future it will also be applied to field.
 - To opt in to applying to both value parameter and field, add '-Xannotation-default-target=param-property' to your compiler arguments.
 - To keep applying to the value parameter only, use the '@param:' annotation target."
 
https://youtrack.jetbrains.com/issue/KT-73255

## PR Checklist

### DIFF
--- a/lapis/src/main/kotlin/org/genspectrum/lapis/response/LapisResponse.kt
+++ b/lapis/src/main/kotlin/org/genspectrum/lapis/response/LapisResponse.kt
@@ -24,29 +24,29 @@ private const val REPORT_TO =
 
 @Schema(description = LAPIS_INFO_DESCRIPTION)
 data class LapisInfo(
-    @Schema(
+    @field:Schema(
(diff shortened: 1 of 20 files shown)


lapis/src/main/kotlin/org/genspectrum/lapis/response/LapisResponse.kt
```diff
@@ -24,29 +24,29 @@ private const val REPORT_TO =
 
 @Schema(description = LAPIS_INFO_DESCRIPTION)
 data class LapisInfo(
-    @Schema(
+    @field:Schema(
```
anna-parker, line 27: so these will now only be applied to the field?

fengelniederhammer: Yes.

PR #1295, opened 2025-08-04
feat(lapis): support TSV with escaped special characters

### LINKED ISSUE #1283: TSV data format without newlines
The default behavior of our TSV output produces for example:

```
column1	column2
some text	"some text
with newline"
another row	...
```

This is a possible format, aligned with the RFC for CSVs, and can be correctly read by programs like Excel. However, as mentioned in #1282 (and discussed in Slack), some libraries/programs/scripts don't support TSV files that contain cells with quoted newlines. To support these programs, we can add an additional data format that transforms the newlines and tabs, so that the output looks like this:

```
column1	column2
some text	some text\nwith newline
another row	...
```

fhennig:
resolves #1283 

Adds a new `IanaTsvWriter`, which escapes new lines and delimiters in cell values, and doesn't do any quoting. Configuring the `CsvPrinter` used in the `CsvWriter` - instead of creating a new class - didn't work: Setting the `QuoteMode`, to `NONE` quotes newlines and delimiters, but the tab was quoted as `\\\t`, which isn't right. I tried working around this, but in the end decided it was cleaner to write a quick implementation of the IANA spec from scratch, instead of trying to get the Apache lib to work with the format which it wasn't designed to support.

I've called the new type "TSV-ESCAPED". The RFC4180 already calls the quote-mechanism escaping, but I'd argue that's not escaping, that's quoting. Instead the new behaviour is escaping. This is now a new datatype, like CSV without headers, to keep things simple. In theory we could have 3 separate settings: TSV/CSV; headers/no headers; quoting/escaping.

## Testing
- I've added a test for the writer itself, to check that it escapes things
- I've added a test to check that the endpoint returns things correctly
- I didn't check it against any data manually, because I don't think the covid data has any cells with newlines, and I'm not sure how to set things up locally with test data.

## PR Checklist

### DIFF
--- a/lapis/src/main/kotlin/org/genspectrum/lapis/response/CsvWriter.kt
+++ b/lapis/src/main/kotlin/org/genspectrum/lapis/response/CsvWriter.kt
@@ -3,6 +3,7 @@ package org.genspectrum.lapis.response
 import org.apache.commons.csv.CSVFormat
 import org.apache.commons.csv.CSVPrinter
 import org.springframework.stereotype.Component
+import java.io.Flushable
 import java.util.stream.Stream
 
 interface RecordCollection<T> {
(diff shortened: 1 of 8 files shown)


lapis/src/main/kotlin/org/genspectrum/lapis/response/CsvWriter.kt
```diff
@@ -44,6 +45,63 @@ class CsvWriter {
...
+    ): String =
+        record.joinToString(separator = delimiter.value.toString()) { cell ->
+            cell
+                .replace("\n", "\\n")
+                .let { v ->
+                    if (delimiter.value == '\t') {
+                        v.replace("\t", "\\t")
+                    } else {
+                        v.replace(delimiter.value.toString(), "\\${delimiter.value}")
+                    }
```
chaoran-chen, line 100: Why do we need two cases here? Wouldn't the `else` case be enough?

lapis/src/test/kotlin/org/genspectrum/lapis/response/CsvWriterTest.kt
```diff
@@ -0,0 +1,74 @@
...
+        private val headers: List<String>,
+        override val records: Stream<List<String?>>,
+    ) : RecordCollection<List<String?>> {
+        override fun getHeader(): List<String> = headers
+
+        override fun mapToCsvValuesList(value: List<String?>): List<String?> = value
+    }
+
+    @Test
+    fun `GIVEN data with nulls and special chars THEN output quotes fields`() {
```
chaoran-chen, line 24:
```suggestion
    fun `GIVEN data with nulls and special chars WHEN using the CsvWriter THEN output quotes fields`() {
```

chaoran-chen (approved):
Looks good to me, thanks!

lapis/src/main/kotlin/org/genspectrum/lapis/response/CsvWriter.kt
fhennig: no because then it will replace \t with \\\t unfortunately

PR #1336, opened 2025-08-21
fix(lapis): fix response schema of /phyloSubtree in OpenAPI spec

fengelniederhammer:
It always returns a single, plain string, but the docs claimed that it would be the usual response object that the other endpoints return. Also since it's a stream with a single entry (as stated in https://github.com/GenSpectrum/LAPIS-SILO/blob/9988e227e808db769961a5c84f733ce33fc250f8/documentation/phylogenetic_queries.md#phylosubtree), we don't need streaming logic, we can simply return the first element of the stream.

Previous response schema in Swagger: 
image

Now:
image

## PR Checklist


lapis/src/main/kotlin/org/genspectrum/lapis/controller/LapisController.kt
chaoran-chen, line 1: Could we set a better example? I suggest

```
(seq1:3,(seq2:2,seq3:1)node2:1)node1:1;
```

chaoran-chen (approved):
Looks good, thanks!

lapis/src/main/kotlin/org/genspectrum/lapis/controller/LapisController.kt
fengelniederhammer: I'll do that in a separate PR. So far we also didn't have a proper example.

PR #1378, opened 2025-10-06
feat: adjust test data to SILO 0.8

### LINKED ISSUE #1300: Update test data for SILO 0.8
SILO 0.8 will introduce a breaking change for the input data. This will require an update to the new format in this repository. After the input data has been adapted, the pinned `SILO_TAG` in https://github.com/GenSpectrum/LAPIS/blob/ebaa1a0f8f1994f99962d4954a63d09eb45113d7/.github/workflows/lapis.yml#L131 should be changed back to `latest`. Also a new entry in the compatibility table should be added

Taepper:
resolves #1300

These files were transformed using the tranformation script in SILO:

```
cat input_file.ndjson |  /path/to/legacyNdjsonTransformer/target/release/legacy-ndjson-transformer > tmp
mv tmp input_file.ndjson
```
in the respective testset directories


fengelniederhammer:
Please note the release note preview. This should not have been a `feat` commit, we don't need changes in the test data in the release notes.

Taepper:
I scoped it to (lapis-e2e). What would have been correct?

fengelniederhammer:
I think I would have labelled it as `chore`. In any case it should not trigger a release.

PR #1393, opened 2025-10-07
fix(silo): close connection to SILO if client cancels request to LAPIS

anna-parker:
resolves #

## PR Checklist

### DIFF
--- a/lapis/build.gradle
+++ b/lapis/build.gradle
@@ -66,6 +66,11 @@ tasks.named('compileTestKotlin') {
     mustRunAfter("generateTestGrammarSource")
 }
 
+tasks.named('runKtlintCheckOverTestSourceSet') {
+    mustRunAfter("generateGrammarSource")
+    mustRunAfter("generateTestGrammarSource")
+}
+
 tasks.named('test') {
     dependsOn generateGrammarSource
     useJUnitPlatform()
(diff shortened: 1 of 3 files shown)


lapis/src/main/kotlin/org/genspectrum/lapis/response/CsvWriter.kt
```diff
@@ -38,8 +39,15 @@ class CsvWriter {
...
-            for (csvRecord in data.getCsvRecords()) {
-                it.printRecord(csvRecord)
+            try {
+                data.getCsvRecords().use { csvRecordStream ->
+                    csvRecordStream.forEach { csvRecord ->
+                        it.printRecord(csvRecord)
+                    }
+                }
+            } catch (e: Exception) {
+                log.error(e) { "Error writing Iana CSV/TSV data" }
```
fengelniederhammer, line 49: One of the two log messages in here should be distinct from the other.

lapis/src/main/kotlin/org/genspectrum/lapis/response/CsvWriter.kt
```diff
@@ -68,8 +76,15 @@ class IanaTsvWriter {
...
         }
-        data.getCsvRecords().forEach { record ->
-            writeRow(appendable, record.map { it.orEmpty() }, delimiter)
+        data.getCsvRecords().use { csvRecordStream ->
+            try {
+                csvRecordStream.forEach { csvRecord ->
+                    writeRow(appendable, csvRecord.map { it.orEmpty() }, delimiter)
+                }
+            } catch (e: Exception) {
+                log.error(e) { "Error writing Iana CSV/TSV data" }
```
fengelniederhammer, line 85: I wonder whether we really want to log at error here. This usually indicates that something is wrong with the code and that we should look at it.

But actually this is a case that simply happens because a client can cause it deliberately (same as sending a malformed request). I think debug (or maybe info) should be enough?

anna-parker: I wrote a todo for this - if the exception is an IOException this is most likely a client disconnected and it should be logged differently otherwise I don't think we expect an error and should log the error in full

PR #1426, opened 2025-11-11
feat(lapis): support random with seed

### LINKED ISSUE #1412: Random with seed
To allow users to get the same random subsample of metadata (through the details endpoint) and sequences, it would be useful if one can set a seed for the randomization. This is already possible in SILO.

A possible syntax could be `orderBy=random(<seed>)`.

This would resolve #602.

fhennig:
resolves #1412 

## Summary

This PR implements deterministic random ordering for LAPIS queries, supporting both unseeded and seeded randomization via GET and POST requests.

* **GET support** for `orderBy=random` and `orderBy=random(<seed>)`. You can still add other `orderBy` fields, but they will be ignored (this is for backwards compatibility).
* **POST support** for:

  * `orderBy: { "random": true }` (unseeded)
  * `orderBy: { "random": <int> }` (seeded)
  * Existing field-based ordering remains supported. Also, mixing is still supported (backwards compatible).

### Internal Changes

* Introduces `OrderBySpec` class, which replaces the `List<OrderByField>`:

  * `ByFields(List<OrderByField>)`
  * `Random(seed: Int?)`

* Ensures that randomized queries are excluded from caching, but randomized queries with seed are cached.

### Testing

* Unit tests for GET and POST parsing, validation, and serialization.
* Integration tests verifying deterministic behaviour (same seed → same order).
* Test for form-encoded
* Test that seeded random is cached

## PR Checklist

### DIFF
--- a/lapis/src/test/kotlin/org/genspectrum/lapis/controller/LapisControllerCommonFieldsTest.kt
+++ b/lapis/src/test/kotlin/org/genspectrum/lapis/controller/LapisControllerCommonFieldsTest.kt
@@ -332,6 +332,28 @@ class LapisControllerCommonFieldsTest(
     }
 
     @Test
+    fun `POST details with random orderBy (form encoded)`() {
+        every {
+            siloQueryModelMock.getDetails(
+                match {
+                    it.orderByFields == OrderBySpec.Random(seed = 123) &&
+                        it.fields == listOf(Field("country"))
+                },
+            )
+        } returns Stream.of(DetailsData(mapOf("country" to TextNode("Switzerland"))))
+
+        val request = postSample(DETAILS_ROUTE)
+            .param("orderBy", "random(123)")
+            .param("orderBy", "age")
+            .param("fields", "country")
+            .contentType(APPLICATION_FORM_URLENCODED)
+
+        mockMvc.perform(request)
+            .andExpect(status().isOk)
+            .andExpect(jsonPath("\$.data[0].country").exists())
+    }
+
+    @Test
     fun `GET aggregated with limit`() {
         every {
             siloQueryModelMock.getAggregated(


fhennig:
requires this: https://github.com/GenSpectrum/LAPIS/pull/1445

lapis/src/main/kotlin/org/genspectrum/lapis/request/OrderByField.kt
```diff
@@ -57,11 +130,33 @@ class OrderByFieldDeserializer(
...
 @Component
 class OrderByFieldConverter(
     private val orderByFieldsCleaner: OrderByFieldsCleaner,
 ) : Converter<String, OrderByField> {
-    override fun convert(source: String) = OrderByField(orderByFieldsCleaner.clean(source), Order.ASCENDING)
+    override fun convert(source: String): OrderByField {
+        val field =
+            if (source.startsWith("random")) {
+                // Validate format: must be "random" or "random(<digits>)"
+                val validRandomPattern = Regex("^random(\\(\\d+\\))?$")
```
fengelniederhammer: This is the same regex as below, isn't it?

lapis/src/main/kotlin/org/genspectrum/lapis/request/OrderByField.kt
```diff
@@ -13,6 +14,78 @@ import org.springframework.boot.jackson.JsonComponent
...
+ * Deserializes the `orderByField`s. Supports a list of fields like:
+ * `[{field: country}, {field: date}]`
+ * as well as random:
+ * `{random: true}`
+ * or random with seed:
+ * `{random: 123}`
+ */
+@JsonComponent
+class OrderBySpecDeserializer(
+    private val orderByFieldsCleaner: OrderByFieldsCleaner,
```
fengelniederhammer: `orderByFieldsCleaner` seems unused.

lapis/src/main/kotlin/org/genspectrum/lapis/request/OrderByField.kt
fengelniederhammer: I find it a bit weird that we need to parse it twice with the same regex.

lapis/src/main/kotlin/org/genspectrum/lapis/model/SiloQueryModel.kt
```diff
@@ -248,7 +250,10 @@ class SiloQueryModel(
                     type = sequenceType,
                     sequenceNames = mapSequenceNames(cleanedSequenceNames, sequenceType),
                     additionalFields = fastaHeaderTemplate.metadataFieldNames,
-                    orderByFields = mapSequenceOrderByFields(sequenceFilters.orderByFields, sequenceType),
+                    orderByFields = mapSequenceOrderByFields(
+                        sequenceFilters.orderByFields.toOrderByFields(),
+                        sequenceType,
+                    ).toOrderBySpec(),
```
fengelniederhammer: This seems odd - why do we need to convert to the list and then back to the spec?

lapis/src/test/kotlin/org/genspectrum/lapis/silo/SiloQueryTest.kt
```diff
@@ -62,6 +64,89 @@ class SiloQueryTest {
         assertThat(objectMapper.readTree(result), equalTo(objectMapper.readTree(expected)))
     }
 
+    @Test
+    fun `GIVEN OrderBySpec Random without seed THEN serializes with randomize true`() {
```
fengelniederhammer: Is there a reason why you didn't put those tests into the `MethodSource` of the `SiloAction is correctly serialized to JSON` test above?

lapis-e2e/test/details.spec.ts
```diff
@@ -306,12 +306,68 @@ key_1002052
...
+
+  it('should order by random with seed', async () => {
+    const result = await lapisClient.postDetails({
+      detailsPostRequest: {
+        orderBy: [{ field: 'random(123)' }],
+        fields: ['primaryKey', 'division'],
+      },
+    });
+
+    expect(result).to.have.nested.property('data[0].division');
```
fengelniederhammer: I suppose we can actually check the values here?

lapis-e2e/test/details.spec.ts
fengelniederhammer: (We could move some of these tests to the `LapisControllerCommonFieldsTest`, I don't know whether that makes them easier. This is the place where I would have expected them but I'm not saying they necessarily need to be there)

lapis/src/test/kotlin/org/genspectrum/lapis/request/OrderByFieldConverterTest.kt
fengelniederhammer: Those tests are nice, but they are missing the integration of the converters etc. with Spring. I didn't find a test that makes sure that an actual GET request is processed correctly, or did I miss it?

lapis/src/main/kotlin/org/genspectrum/lapis/openApi/OpenApiDocs.kt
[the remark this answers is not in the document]
fhennig: Updated to: "The fields of the response to order by. Fields specified here must also be present in "fields". You can also use "random" or "random(SEED)" where SEED is an integer."

lapis/src/test/kotlin/org/genspectrum/lapis/silo/SiloQueryTest.kt
fhennig: no reason. Fixed it.

Issue #1423, opened 2025-11-06
CI: zstd compression via header doesn't work


fhennig:
example:

```
   9) All endpoints
       (/alignedAminoAcidSequences/S)
         should return zstd compressed data when accepting compression in header:
     Error: AssertionError: expected [ 62, 107, 101, 121 ] to deeply equal [ 40, 181, 47, 253 ]
ArrayBuffer content (for context): >key_1408408
MFVFLVLLPLVSSQCVNLTTRTQLPPAYTNSFTRGVYYPDKVFRSSVLHSTQDLFLPFFSNVTWFHAI--SGTNGTKRFDNPVLPFNDGVYFASTEKSNIIRGWIFGTTLDSKTQSLLIVNNATNVVIKVCEFQFCNDPFLGV-YHKNNKSWMESEFRVYSSANNCTFEYVSQPFLMDLEGKQGNFKNLREFVFKNIDGYFKIYSKHTPINLVRDLPQGFSALEPLVDLPIGINITRFQTLLALHRSYLTPGDSSSGWTAGAAAYYVGYLQPRTFLLKYNENGTITD...
      at expectIsZstdEncoded (test/common.ts:94:11)
      at Context.<anonymous> (test/common.spec.ts:222:9)
      at process.processTicksAndRejections (node:internal/process/task_queues:105:5)

```

all the tests with "should return zstd compressed data when accepting compression in header" are failing in CI recently.

This commit on main has a green test: https://github.com/GenSpectrum/LAPIS/commit/f911c7106c787795a1b1bd7c01e95cb45b6aef54

I tried with LAPIS 0.6.1 and the same SILO version that was used back then (`commit-a1dcc46e52bd5e9d439be49b1acbb50a6898aeb0`) but I still get the failure: https://github.com/GenSpectrum/LAPIS/actions/runs/19143938994/job/54716439848?pr=1422

I don't get it locally at all, I tried a few different version combinations.

More CI tests I did in this PR: https://github.com/GenSpectrum/LAPIS/pull/1422 -- It also includes the little snippet which prints the clear text (see above error message)

---

I've made a PR to disable the test in CI for now, until we can sort this out: https://github.com/GenSpectrum/LAPIS/pull/1431

fhennig:
I spent 3 hours trying to debug this last week, but couldn't do it. Maybe @fengelniederhammer we can have a look together, maybe on Thursday?

PR #1431, opened 2025-11-13
test(lapis-e2e): skip test that only fails in CI for now

fhennig:
We are tracking this in #1423

For now, we need to disable this test, because it's too noisy. we can re-enable it once we have fixed it.


fhennig:
I'm running another PR to check about the remaining failures: https://github.com/GenSpectrum/LAPIS/pull/1432

PR #1499, opened 2026-01-12
feat(lapis): remove `protected` OpennessLevel

### LINKED ISSUE #1498: Remove the protected OpennessLevel
We don't need the `OpennessLevel.PROTECTED` mode anymore. (A proper authorization framework will be implemented through #251.)

fengelniederhammer:
resolves #1498

I still left the `opennessLevel` field in the database config, since we might need it we implement proper authorization. Or do you think we should remove it now?

I also didn't label this as breaking (althought it technically is) since noone was using it and it was an intentionally undocumented feature. Or should we even mark it as `chore` so that it doesn't show up in the changelog?

## PR Checklist


fhennig (approved):
LGTM!

I think 'feat' is fine, and also making it non-breaking. This way we can find it in the changelog, but yes, it's undocumented so it's not breaking, fine for me.

chaoran-chen:
Thanks, sounds all good to me!

PR #1509, opened 2026-01-21
feat(lapis): add overall statistics to .*OverTime endpoints

### LINKED ISSUE #1349: mutationsOverTime/queriesOverTime: supply overall mutation data
In the frontend we need total proportion for a mutation, so we can apply the proportion filter. We now need the same thing for the new queries.

For mutations, this is currently solved with the `FetchSubstitutionsOrDeletionsOperator` which calls the `nucleotideMutations` endpoint (and similarly for AAs). But for the new queries, we don't have an equivalent endpoint.

Something like this would be needed (property name can be different):

```
  {
    queries: [...],
    dateRanges: [...],
    data: [...],           // existing time-series
    totalCountsByDateRange: [...],
    overallStatisticsByQuery: [ // NEW
      { count: 100, coverage: 150, proportion: 0.67 },
      ...
    ]
  }
```

fhennig:
resolves #1349 

## PR Checklist

### DIFF
--- a/lapis-e2e/test/queriesOverTime.spec.ts
+++ b/lapis-e2e/test/queriesOverTime.spec.ts
@@ -41,6 +41,10 @@ describe('The /mutationsOverTime endpoint', () => {
         ],
       ],
       totalCountsByDateRange: [22, 77, 0],
+      overallStatisticsByQuery: [
+        { count: 58, coverage: 95, proportion: 58 / 95 },
+        { count: 1, coverage: 95, proportion: 1 / 95 },
+      ],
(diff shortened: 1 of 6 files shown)


fhennig:
(There is a test failure still)

Hmm, I'm now doubting whether this is actually all that useful. I guess this kind of aggregation could be done in the frontend as well.

What do you think @fengelniederhammer ?

fhennig:
Okay, I've changed my mind, I think this is actually not very useful to implement, since the data is already there.

PR #1532, opened 2026-02-11
feat(lapis): support `filename*` with UTF-8 character encoding

### LINKED ISSUE #1494: Use filename* in Content-Disposition
When defining the filename of a downloaded file, LAPIS currently uses `filename=` in the `Content-Disposition` header. This comes with many restrictions regarding the filename and it would be better to use `filename*=` which has full UTF-8 support (see [RFC 5987](https://datatracker.ietf.org/doc/html/rfc5987#section-3.2) and [MDN](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Content-Disposition)).

fhennig:
resolves #1494 

- Use Springs built-in `ContentDisposition` class to generate the correct `filename*` header
- Use only ASCII chars for the `filename` header (See also https://github.com/spring-projects/spring-framework/issues/31940 - The `ContentDisposition` class will encode `filename` as well, we do not want that because not all clients will decode headers)
- Put all of the code into a dedicated utility file
- Add tests accordingly

## PR Checklist

### DIFF
--- a/lapis/src/test/kotlin/org/genspectrum/lapis/util/ContentDispositionUtilsTest.kt
+++ b/lapis/src/test/kotlin/org/genspectrum/lapis/util/ContentDispositionUtilsTest.kt
@@ -58,6 +58,11 @@ class ContentDispositionUtilsTest {
                     "data.json.gz",
                     "attachment; filename=data.json.gz; filename*=UTF-8''data.json.gz",
                 ),
+                Arguments.of(
+                    "filename with semicolon",
+                    "file;name.json",
+                    "attachment; filename=filename.json; filename*=UTF-8''file%3Bname.json",
+                ),
             )
 
         @JvmStatic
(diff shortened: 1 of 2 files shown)


lapis/src/test/kotlin/org/genspectrum/lapis/util/ContentDispositionUtilsTest.kt
```diff
@@ -0,0 +1,77 @@
...
+            listOf(
+                Arguments.of("ASCII unchanged", "test.json", "test.json"),
+                Arguments.of("Remove accents", "données.json", "donnes.json"),
+                Arguments.of("Remove Chinese", "文件.txt", ".txt"),
+                Arguments.of("Preserve space", "my file.csv", "my file.csv"),
+                Arguments.of("Mixed content", "file_测试_2024.json", "file__2024.json"),
+                Arguments.of("Remove double quotes", "file\"name.json", "filename.json"),
+                Arguments.of("Remove carriage return", "file\rname.json", "filename.json"),
+                Arguments.of("Remove newline", "file\nname.json", "filename.json"),
+                Arguments.of("Remove all three", "file\"\r\nname.json", "filename.json"),
```
fengelniederhammer: Let's add a test case for `;`. I think that did the most harm when I tested it, because it splits the filename. (`\n` and `\r` didn't make it to the response, I'm guessing that Spring or Tomcat filters those out it header values)

PR #1558, opened 2026-02-24
feat(lapis-e2e): add more testing for advanced variant queries

fhennig:
The main reason for adding this, is `count_and_coverage_queries.json` - which documents how we can construct a coverage query from a count query, as well as serving as a regression test for the future. This is related to the "queries over time" feature, where we need a count and coverage query to calculate the proportion of matching sequences, when general coverage in the sequences is quite low and we therefore explicitly want to count the sequences that do actually provide coverage for the count query.

The key thing is that for a count query `q`, the coverage can be determined with `q or not maybe(q)`, which can be read as "either we have a direct count hit, or we definitely know that it cannot be a hit." This is not the same as simply having coverage for all mentioned positions (this is also shown in the test case). We can have only partial coverage, but still a valid query hit.

## Implementation

- Create variantQueries.spec.ts with support for multiple variant queries per test case
- Add JSON-driven test case format allowing sequential requests with expected responses
- Add relevant test cases
- Add documentation as READMEs in the relevant dirs
- Fix QueryControllerApi import in common.ts

Generated with [Claude Code](https://claude.com/claude-code) and polished by Felix

## PR Checklist

### DIFF
--- a/lapis-e2e/test/advancedVariantQueries.spec.ts
+++ b/lapis-e2e/test/advancedVariantQueries.spec.ts
@@ -32,8 +32,6 @@ function filterUndefinedValues(response: AggregatedResponse): AggregatedResponse
 }
 
 describe('Advanced Variant Query Tests', function () {
-  this.timeout(10000); // Set timeout to 10 seconds for all tests in this suite
-
   advancedQueryFiles
     .map(file => JSON.parse(fs.readFileSync(`${queriesPath}/${file}`).toString()))
     .forEach((testCase: VariantQueryTestCase) => {


Taepper:
Excellent solution and best use-case for the maybe construct!

Taepper (commented):
The tests look sensible to me, but maybe @fengelniederhammer can judge it better in context to the remaining LAPIS sources / tests

Taepper (approved):

fengelniederhammer (approved):

PR #1593, opened 2026-03-12
fix(lapis): set a low timeout on calling SILO info so that the health endpoint doesn't block if SILO doesn't respond

fengelniederhammer:
I tested it locally by start SILO locally with a "sleep one minute" in the info endpoint.

## PR Checklist

### DIFF
--- a/lapis/src/main/kotlin/org/genspectrum/lapis/silo/SiloClient.kt
+++ b/lapis/src/main/kotlin/org/genspectrum/lapis/silo/SiloClient.kt
@@ -23,6 +23,7 @@ import java.net.http.HttpClient
 import java.net.http.HttpRequest
 import java.net.http.HttpResponse
 import java.net.http.HttpResponse.BodyHandlers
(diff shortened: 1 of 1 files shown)


fhennig (approved):
sounds good to me!

corneliusroemer (commented):
This PR caused new flakes on Loculus CI because sometimes the info call does take longer than 100ms due to high load.

I understand that the LAPIS health endpoint should not stall when SILO blocks - but shouldn't this be handled in the health endpoint's call to this function here rather than the function itself?

A side-effect, and I don't think an intended one, is that now all callers to `/sample/info` get a 503 when this call takes >100ms. But there's nothing intrinsically wrong about taking >100ms - it can happen under load. Callers can set their own timeouts however they like, some might be happy to wait for a second but now they get a 503 instead.

Maybe moving where the timeout lives (to the caller that wants it) is better than #1866 (which is only a narrow workaround for the underlying problem - if my understanding above is correct).

PR #1626, opened 2026-03-31
feat(lapis): also allow multiple value in int, float and date filters (and concatenate them with `or`)

### LINKED ISSUE #1595: Search for Int and float fields only allows exact match
When calling the LAPIS aggregated and details endpoint LAPIS enforces (see the `extractSingleFilterValue` check) that number fields only contain one value, whereas string fields can be configured to search for multiple values. 

See thread in Loculus: https://loculus.slack.com/archives/C05G172HL6L/p1773316824073099 - it would be nice to enable this functionality for number fields as well

fengelniederhammer:
resolves #1595

`int`, `float`, and `date` exact-match filters now accept an array of values, combining them with logical OR — the same behaviour already supported for `string` fields.

```json
{ "age": [50, 51, 52] }
{ "qcValue": [0.95, 0.97, 0.8] }
{ "date": ["2021-05-08", "2021-01-20"] }
```

`null` is supported as an array element to filter for missing values. Range filters (`*From`/`*To`) are unchanged and still accept a single value only.

## PR Checklist

### DIFF
--- a/lapis-e2e/test/aminoAcidSequence.spec.ts
+++ b/lapis-e2e/test/aminoAcidSequence.spec.ts
@@ -190,7 +190,7 @@ describe('The /alignedAminoAcidSequence endpoint', () => {
 
       const errorResponse = await response.json();
       expect(errorResponse.error.detail).to.match(
-        /Error from SILO: The table does not contain the SequenceColumn 'unknownGene'/
+        /Error from SILO: The table does not contain the field unknownGene/
       );
     });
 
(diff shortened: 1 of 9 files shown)


lapis-e2e/test/aminoAcidSequence.spec.ts
```diff
@@ -190,7 +190,7 @@ describe('The /alignedAminoAcidSequence endpoint', () => {
 
       const errorResponse = await response.json();
       expect(errorResponse.error.detail).to.match(
-        /Error from SILO: The table does not contain the SequenceColumn 'unknownGene'/
+        /Error from SILO: The table does not contain the field unknownGene/
```
fengelniederhammer, line 193: Unrelated to this PR, but SILO recently changed the error message in this case.

fhennig: ok

fhennig (approved):
LGTM, but the `null` documentation case seems worth attemting to fix.

PR #1650, opened 2026-04-20
feat(lapis): fetch data from SILO in the Apache Arrow format

### LINKED ISSUE #1547: Get Arrow data from SILO
After https://github.com/GenSpectrum/LAPIS-SILO/pull/1164, LAPIS should get data from SILO using the Arrow format.

fengelniederhammer:
resolves #1547

SILO supports returning query responses as [Apache Arrow IPC streams](https://arrow.apache.org/docs/format/Columnar.html#format-ipc) (`application/vnd.apache.arrow.stream`), which is more efficient to parse than NDJSON. This PR switches LAPIS to request and consume Arrow IPC responses from SILO for all query types.

**For reviewers:**
- `SiloClient.kt`: sends `Accept: application/vnd.apache.arrow.stream`, reads the response as a lazy `Stream` via `ArrowStreamReader`. A shared `RootAllocator` bean (Netty-backed) is used application-wide; a child allocator is created per query and closed when the stream is consumed.
- `SiloQuery.kt`: each `SiloAction` subclass now carries an `arrowConverter` reference instead of a Jackson `TypeReference`.
- The JVM flags `-Dio.netty.tryReflectionSetAccessible=true` and `--add-opens=java.base/java.nio=ALL-UNNAMED` are required and added to `bootRun`, `test`, and `entrypoint.sh`.

## PR Checklist

### DIFF
--- a/.idea/runConfigurations/LapisMultiSegmented.xml
+++ b/.idea/runConfigurations/LapisMultiSegmented.xml
@@ -3,6 +3,7 @@
     <module name="lapis.main" />
     <option name="PROGRAM_PARAMETERS" value="--silo.url=http://localhost:8093 --lapis.databaseConfig.path=lapis-e2e/testData/multiSegmented/testDatabaseConfig.yaml --referenceGenomeFilename=lapis-e2e/testData/multiSegmented/reference_genomes.json --server.port=8094" />
     <option name="SPRING_BOOT_MAIN_CLASS" value="org.genspectrum.lapis.LapisApplicationKt" />
+    <option name="VM_PARAMETERS" value="-Dio.netty.tryReflectionSetAccessible=true --add-opens=java.base/java.nio=ALL-UNNAMED" />
     <option name="WORKING_DIRECTORY" value="file://$PROJECT_DIR$" />
     <method v="2">
       <option name="Make" enabled="true" />
--- a/lapis/src/main/kotlin/org/genspectrum/lapis/silo/ArrowRowConverter.kt
+++ b/lapis/src/main/kotlin/org/genspectrum/lapis/silo/ArrowRowConverter.kt
@@ -25,11 +25,15 @@ import org.genspectrum.lapis.response.PhyloSubtreeData
 import org.genspectrum.lapis.response.SequenceData
 import org.genspectrum.lapis.util.UNALIGNED_PREFIX
 
-/** Converts one row at `rowIndex` from an Arrow [org.apache.arrow.vector.VectorSchemaRoot] to a typed Kotlin object. */
+/**
+ * Converts one row at `rowIndex` from an Arrow [org.apache.arrow.vector.VectorSchemaRoot] to a typed Kotlin object.
+ *
+ * See https://github.com/GenSpectrum/LAPIS-SILO/blob/main/documentation/query_documentation.md#action for the SILO response schema
+ */
 typealias ArrowRowConverter<T> = (root: VectorSchemaRoot, rowIndex: Int) -> T
 
 val AGGREGATION_DATA_ARROW_CONVERTER: ArrowRowConverter<AggregationData> = { root, rowIndex ->
-    val count = root.getLong(COUNT_PROPERTY, rowIndex).toInt()
+    val count = root.getLong(COUNT_PROPERTY, rowIndex)
     val fields = root.schema.fields
         .mapIndexedNotNull { colIdx, field ->
             if (field.name == COUNT_PROPERTY) {
--- a/lapis/src/main/kotlin/org/genspectrum/lapis/silo/ArrowRowConverter.kt
+++ b/lapis/src/main/kotlin/org/genspectrum/lapis/silo/ArrowRowConverter.kt
@@ -1,0 +1,159 @@
+package org.genspectrum.lapis.silo
+
+import com.fasterxml.jackson.databind.JsonNode
+import com.fasterxml.jackson.databind.node.BooleanNode
+import com.fasterxml.jackson.databind.node.DoubleNode
+import com.fasterxml.jackson.databind.node.FloatNode
+import com.fasterxml.jackson.databind.node.IntNode
+import com.fasterxml.jackson.databind.node.LongNode
+import com.fasterxml.jackson.databind.node.NullNode
+import com.fasterxml.jackson.databind.node.TextNode
+import org.apache.arrow.vector.BigIntVector
+import org.apache.arrow.vector.BitVector
+import org.apache.arrow.vector.Float4Vector
+import org.apache.arrow.vector.Float8Vector
+import org.apache.arrow.vector.IntVector
+import org.apache.arrow.vector.VarCharVector
+import org.apache.arrow.vector.VectorSchemaRoot
+import org.genspectrum.lapis.response.AggregationData
+import org.genspectrum.lapis.response.COUNT_PROPERTY
+import org.genspectrum.lapis.response.DetailsData
+import org.genspectrum.lapis.response.InsertionData
+import org.genspectrum.lapis.response.MostCommonAncestorData
+import org.genspectrum.lapis.response.MutationData
+import org.genspectrum.lapis.response.PhyloSubtreeData
+import org.genspectrum.lapis.response.SequenceData
+import org.genspectrum.lapis.util.UNALIGNED_PREFIX
+
+/** Converts one row at `rowIndex` from an Arrow [org.apache.arrow.vector.VectorSchemaRoot] to a typed Kotlin object. */
+typealias ArrowRowConverter<T> = (root: VectorSchemaRoot, rowIndex: Int) -> T
+
+val AGGREGATION_DATA_ARROW_CONVERTER: ArrowRowConverter<AggregationData> = { root, rowIndex ->
+    val count = root.getLong(COUNT_PROPERTY, rowIndex)?.toInt() ?: 0
+    val fields = root.schema.fields
+        .filter { it.name != COUNT_PROPERTY }
+        .mapIndexed { _, field ->
+            val actualColIdx = root.schema.fields.indexOfFirst { it.name == field.name }
+            field.name to root.fieldValueAsJsonNode(actualColIdx, rowIndex)
+        }
+        .toMap()
+    AggregationData(count, fields)
+}
+
+val DETAILS_DATA_ARROW_CONVERTER: ArrowRowConverter<DetailsData> = { root, rowIndex ->
+    DetailsData(
+        root.schema.fields
+            .mapIndexed { colIdx, field -> field.name to root.fieldValueAsJsonNode(colIdx, rowIndex) }
+            .toMap(),
+    )
+}
+
+val MOST_COMMON_ANCESTOR_DATA_ARROW_CONVERTER: ArrowRowConverter<MostCommonAncestorData> = { root, rowIndex ->
+    MostCommonAncestorData(
+        mrcaNode = root.getString("mrcaNode", rowIndex),
+        missingNodeCount = root.getInt("missingNodeCount", rowIndex) ?: 0,
+        missingFromTree = root.getString("missingFromTree", rowIndex),
+    )
+}
+
+val MUTATION_DATA_ARROW_CONVERTER: ArrowRowConverter<MutationData> = { root, rowIndex ->
+    MutationData(
+        mutation = root.getString("mutation", rowIndex),
+        count = root.getInt("count", rowIndex),
+        proportion = root.getDouble("proportion", rowIndex),
+        sequenceName = root.getString("sequenceName", rowIndex),
+        mutationFrom = root.getString("mutationFrom", rowIndex),
+        mutationTo = root.getString("mutationTo", rowIndex),
+        position = root.getInt("position", rowIndex),
+        coverage = root.getInt("coverage", rowIndex),
+    )
+}
+
+val INSERTION_DATA_ARROW_CONVERTER: ArrowRowConverter<InsertionData> = { root, rowIndex ->
+    InsertionData(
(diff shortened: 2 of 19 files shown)


lapis/src/test/kotlin/org/genspectrum/lapis/silo/ArrowTestHelper.kt
```diff
@@ -0,0 +1,139 @@
...
+fun buildArrowIpcStream(rows: List<Map<String, Any?>>): ByteArray {
+    if (rows.isEmpty()) {
+        return buildEmptyArrowIpcStream()
+    }
+
+    val sampleRow = rows.first()
+    val allocator = RootAllocator()
+
+    val fields = sampleRow.entries.map { (name, value) ->
+        when (value) {
```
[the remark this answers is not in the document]
fengelniederhammer: It's only for tests, this should be fine.

fengelniederhammer:
I tested this on staging with the Covid dataset
* clicking through the dashboards showed nothing suspicious
  * The nucleotide mutations over time had trouble loading within the timeout of 1 minute, but that's also already the case on current prod
* I downloaded several million sequences (with zstd compression), it worked. Also 3 of these downloads worked.

fengelniederhammer:
## AI test performed against staging

Tested against https://gs-staging-1.int.genspectrum.org/open/v2/ on live dataset (~9.4M sequences).

| Test | Result |
|---|---|
| Aggregated total count | 9,425,705 ✓ |
| Aggregated by country | 134 countries, counts sane ✓ |
| Aggregated multi-field (country × pangoLineage) | 24,335 combos ✓ |
| Details — field names, types, nulls | 54 fields, int/float/null correct ✓ |
| Nucleotide mutations | strings, proportions, positions correct ✓ |
| Nucleotide insertions | insertion strings intact ✓ |
| Zero-result filter | `count: 0`, no crash ✓ |
| Bad input → error propagation | 400 with detail message ✓ |
| Aligned sequences | 29,903 chars, only `{-, A, C, G, N, T}` ✓ |
| UTF-8 correctness | strain strings round-trip clean ✓ |
| Large response (all mutations, no limit) | 1.8 MB, HTTP 200, no crash ✓ |

All pass. No Arrow parse errors or allocator leak warnings observed.

fengelniederhammer:
I just noticed that there might still be a memory leak somewhere. I'll observe this. LAPIS' memory went up to 1.5 GB but didn't go down yet.

fengelniederhammer:
> I just noticed that there might still be a memory leak somewhere. I'll observe this. LAPIS' memory went up to 1.5 GB but didn't go down yet.

Ok, doesn't seem the case. Memory peaks at 1.7 GB and doesn't increase any further while LAPIS stays operational.

.idea/runConfigurations/LapisMultiSegmented.xml
```diff
@@ -3,6 +3,7 @@
     <module name="lapis.main" />
     <option name="PROGRAM_PARAMETERS" value="--silo.url=http://localhost:8093 --lapis.databaseConfig.path=lapis-e2e/testData/multiSegmented/testDatabaseConfig.yaml --referenceGenomeFilename=lapis-e2e/testData/multiSegmented/reference_genomes.json --server.port=8094" />
     <option name="SPRING_BOOT_MAIN_CLASS" value="org.genspectrum.lapis.LapisApplicationKt" />
+    <option name="VM_PARAMETERS" value="-Dio.netty.tryReflectionSetAccessible=true --add-opens=java.base/java.nio=ALL-UNNAMED" />
```
fhennig, line 6: Here (and in the LapisOpen.xml) can we also add the comment what the VM args are for? Like we did in build.gradle.

lapis/src/main/kotlin/org/genspectrum/lapis/silo/SiloQuery.kt
```diff
@@ -50,7 +33,7 @@ interface CommonActionFields {
 const val ORDER_BY_RANDOM_FIELD_NAME = "random"
 
 sealed class SiloAction<ResponseType>(
-    @JsonIgnore val typeReference: TypeReference<ResponseType>,
```
fhennig, line 53: We'll now always use arrow, there is no fallback to NDJSON, and all the JSON code is gone. So we're sure we want to use arrow forever and not go back, right? 

It kind of says that in the PR - but I still wanted to explicitly bring it up. Did we consider supporting both?

lapis/src/main/kotlin/org/genspectrum/lapis/silo/ArrowRowConverter.kt
```diff
@@ -0,0 +1,186 @@
...
+import org.genspectrum.lapis.response.MutationData
+import org.genspectrum.lapis.response.PhyloSubtreeData
+import org.genspectrum.lapis.response.SequenceData
+import org.genspectrum.lapis.util.UNALIGNED_PREFIX
+
+/** Converts one row at `rowIndex` from an Arrow [org.apache.arrow.vector.VectorSchemaRoot] to a typed Kotlin object. */
+typealias ArrowRowConverter<T> = (root: VectorSchemaRoot, rowIndex: Int) -> T
+
+val AGGREGATION_DATA_ARROW_CONVERTER: ArrowRowConverter<AggregationData> = { root, rowIndex ->
+    val count = root.getLong(COUNT_PROPERTY, rowIndex).toInt()
```
fhennig: Can we link to SILO API docs that document the presence of this field in the SILO response? 

It's fine if we can't (the JSON converters in SiloResponse.kt also didn't have this) but as a reader of this file I would like to link the converters to the SILO API docs.

Even more relevant that the field name maybe is also the type. is `getLong` (bigint) correct? What if SILO sends an int32?

lapis/src/main/kotlin/org/genspectrum/lapis/silo/ArrowRowConverter.kt
```diff
@@ -0,0 +1,186 @@
...
+    InsertionData(
+        count = root.getInt("count", rowIndex),
+        insertion = root.getString("insertion", rowIndex),
+        insertedSymbols = root.getString("insertedSymbols", rowIndex),
+        position = root.getInt("position", rowIndex),
+        sequenceName = root.getString("sequenceName", rowIndex),
+    )
+}
+
+val PHYLO_SUBTREE_DATA_ARROW_CONVERTER: ArrowRowConverter<PhyloSubtreeData> = { root, rowIndex ->
```
fhennig, line 88: It seems like there is no test for this converter - can we add one?

fengelniederhammer:
## Benchmarking request durations (using Arrow vs. not using Arrow for communication with SILO)

<details>

<summary>Script</summary>

```
#!/bin/bash
# Benchmarks Arrow-based staging vs NDJSON-based prod.
# Requires hyperfine: https://github.com/sharkdp/hyperfine

STAGING="https://gs-staging-1.int.genspectrum.org/open/v2"
PROD="https://lapis.cov-spectrum.org/open/v2"

SEQUENCES_CALL="/sample/alignedNucleotideSequences?limit=100000&compression=zstd"
METADATA_CALL="/sample/details?limit=1000000&compression=zstd"

WARMUP=1
RUNS=10

echo "=== Aligned nucleotide sequences ==="
hyperfine --warmup $WARMUP --runs $RUNS --show-output \
    "curl -s -o /dev/null '$STAGING$SEQUENCES_CALL'" \
    "curl -s -o /dev/null '$PROD$SEQUENCES_CALL'"

echo ""
echo "=== Details / metadata ==="
hyperfine --warmup $WARMUP --runs $RUNS \
    "curl -s -o /dev/null '$STAGING$METADATA_CALL'" \
    "curl -s -o /dev/null '$PROD$METADATA_CALL'"
```

</details>

```
=== Aligned nucleotide sequences ===
Benchmark 1: curl -s -o /dev/null 'https://gs-staging-1.int.genspectrum.org/open/v2/sample/alignedNucleotideSequences?limit=100000&compression=zstd'
  Time (mean ± σ):     30.775 s ±  0.043 s    [User: 0.049 s, System: 0.043 s]
  Range (min … max):   30.731 s … 30.844 s    10 runs
 
Benchmark 2: curl -s -o /dev/null 'https://lapis.cov-spectrum.org/open/v2/sample/alignedNucleotideSequences?limit=100000&compression=zstd'
  Time (mean ± σ):     30.915 s ±  0.072 s    [User: 0.078 s, System: 0.071 s]
  Range (min … max):   30.774 s … 30.991 s    10 runs
 
Summary
  curl -s -o /dev/null 'https://gs-staging-1.int.genspectrum.org/open/v2/sample/alignedNucleotideSequences?limit=100000&compression=zstd' ran
    1.00 ± 0.00 times faster than curl -s -o /dev/null 'https://lapis.cov-spectrum.org/open/v2/sample/alignedNucleotideSequences?limit=100000&compression=zstd'

=== Details / metadata ===
Benchmark 1: curl -s -o /dev/null 'https://gs-staging-1.int.genspectrum.org/open/v2/sample/details?limit=1000000&compression=zstd'
  Time (mean ± σ):     18.526 s ±  0.250 s    [User: 1.692 s, System: 1.273 s]
  Range (min … max):   18.125 s … 18.847 s    10 runs
 
Benchmark 2: curl -s -o /dev/null 'https://lapis.cov-spectrum.org/open/v2/sample/details?limit=1000000&compression=zstd'
  Time (mean ± σ):     20.889 s ±  0.091 s    [User: 1.411 s, System: 0.884 s]
  Range (min … max):   20.768 s … 21.001 s    10 runs
 
Summary
  curl -s -o /dev/null 'https://gs-staging-1.int.genspectrum.org/open/v2/sample/details?limit=1000000&compression=zstd' ran
    1.13 ± 0.02 times faster than curl -s -o /dev/null 'https://lapis.cov-spectrum.org/open/v2/sample/details?limit=1000000&compression=zstd'
```

There is only a neglectable difference when downloading sequences.
Downloading metadata however is faster with Arrow.

lapis/src/main/kotlin/org/genspectrum/lapis/silo/ArrowRowConverter.kt
```diff
@@ -0,0 +1,186 @@
...
+        return null
+    }
+    return String(vector.get(rowIndex), Charsets.UTF_8)
+}
+
+private fun VectorSchemaRoot.getString(
+    name: String,
+    rowIndex: Int,
+): String {
+    val vector = getVector(name) as? VarCharVector
```
fhennig: `getVector(name)` is called twice. Use a local variable.

lapis/src/test/kotlin/org/genspectrum/lapis/silo/ArrowTestHelper.kt
```diff
@@ -0,0 +1,139 @@
...
+                        vector.setNull(rowIdx)
+                    } else {
+                        vector.setSafe(rowIdx, (value as String).toByteArray(Charsets.UTF_8))
+                    }
+                }
+            }
+        }
+    }
+
+    root.setRowCount(rows.size)
```
fhennig: This looks like it's duplicated from line 62?

lapis/src/test/kotlin/org/genspectrum/lapis/silo/ArrowTestHelper.kt
```diff
@@ -0,0 +1,139 @@
...
+                name,
+                FieldType.nullable(
+                    ArrowType.FloatingPoint(org.apache.arrow.vector.types.FloatingPointPrecision.DOUBLE),
+                ),
+                null,
+            )
+
+            is Boolean -> Field(name, FieldType.nullable(ArrowType.Bool()), null)
+
+            else -> Field(name, FieldType.nullable(ArrowType.Utf8()), null)
```
fhennig, line 58: Setting the type like this for null fields can lead to incorrectly typed fields - which _maybe_ can cause issues down the line? I see no obvious way to change this, but maybe we can add a comment mentioning this.

fhennig (changes requested):

lapis/src/main/kotlin/org/genspectrum/lapis/silo/SiloQuery.kt
fengelniederhammer: @chaoran-chen Are we sure that we want to stick to the Arrow format? Or do we want to keep the NDJSON format? (In any case we could also revert this PR)

.idea/runConfigurations/LapisMultiSegmented.xml
fengelniederhammer: Those are autogenerated, I'd rather not edit them. I added a section in the README instead.

lapis/src/main/kotlin/org/genspectrum/lapis/silo/ArrowRowConverter.kt
fengelniederhammer: Thinking about it, we can simply leave the count as a long. There is no need to downcast it.

I also added a link to the general arrow converter (rather than to a single instantiation of it)

fengelniederhammer:
I did some more tests (staging with arrow vs prod without arrow):
* I did not observe differences when downloading sequences
  * Downloading uncompressed sequences from both satisfy my bandwidth (~13 MB/s)
  * Downloading compressed sequences from both has around the same download speed (~300 kB/s)
* I did not observe differences when querying mutations over time. I manually sent two requests (making sure they are not cached): On staging they took 24.0 and 24.3 seconds, on prod both took 24.5 seconds.

taepper:
> I did not observe differences when downloading sequences

Could you add a last test that downloads all metadata?

fhennig (approved):
Thanks! LGTM

fengelniederhammer:
I created #1669 for follow up analysis.

PR #1681, opened 2026-05-09
fix(lapis-docs): resolve TypeScript warnings and add astro check to CI

### LINKED ISSUE #534: Docs: Check why `astro check` throws weird errors
`npx astro check` complains about a lot of weird things, like
```
src/components/TabsBox/react/TabsBox.tsx:22:18 - error ts(2322): Type '{ children: any[]; className: string; }' is not assignable to type 'HTMLAttributes'.
  Property 'className' does not exist on type 'HTMLAttributes'.

22             <div className='tabs'>
  
```
Check why this is the case and whether we can run this check in the CI.

fhennig:
Closes #534

## Summary
- Fix unused variable/import warnings (`_` prefix for unused params, remove unused imports)
- Fix `Object is possibly 'null'` error in `TabsBox.astro` with `?? ''` fallback
- Install `@types/swagger-ui-react` to resolve missing declaration file error
- Add `npm run check-astro` (`astro check`) step to the CI `checks` job

### DIFF
--- a/.github/workflows/lapis-docs.yml
+++ b/.github/workflows/lapis-docs.yml
@@ -25,6 +25,7 @@ jobs:
       - run: npm ci
       - run: npm run check-format
       - run: npm run check-types
+      - run: npm run check-astro
 
   dockerImage:
     name: Build Docs Docker Image And Run E2E Tests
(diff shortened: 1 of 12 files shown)


fengelniederhammer (approved):
Thanks!

Issue #1682, opened 2026-05-11
Send SILO queries as SaneQL


fengelniederhammer:
https://github.com/GenSpectrum/LAPIS-SILO/issues/1182 changes the SILO queries to SaneQL. Adapt LAPIS accordingly.

Details remain to be sketched out.

PR #1700, opened 2026-05-21
feat(lapis): also handle SILO returning arrow date values

Taepper:
For testing, we will first need to handle sending saneql queries, which has been finished in #1682 

Now this PR adds handling of returned arrow batches that also contain date arrays

## PR Checklist


fengelniederhammer (commented):
What Copilot said. The `SiloClientTest` tests the deserialization of responses, I think that's were we should also test this change.

Otherwise looks good, the e2e are back to green!

fengelniederhammer (approved):
Thanks! :)

PR #1702, opened 2026-05-23
docs(lapis): improve landing page, describe lineage queries

chaoran-chen:
This is updating the docs, mainly:

- add content to landing page
- describe lineage queries
- describe customizable FASTA headers

## PR Checklist
- ~[ ] All necessary documentation has been adapted.~
- ~[ ] All necessary changes are explained in the `llms.txt`.~
- ~[ ] The implemented feature is covered by an appropriate test.~

### DIFF
--- a/lapis-docs/src/content/docs/concepts/customizable-fasta-headers.mdx
+++ b/lapis-docs/src/content/docs/concepts/customizable-fasta-headers.mdx
@@ -5,7 +5,7 @@ description: Customize the header line of FASTA records returned by sequence end
 
 When requesting sequences as FASTA, you can customize the header line of each record by setting the
 `fastaHeaderTemplate` request property.
-The parameter is ignored if the data format is not FASTA.
+An error is thrown if the prameter is set but the data format is not FASTA.
 
 The template allows placeholders in curly braces:
 
--- a/lapis-docs/src/content/docs/concepts/customizable-fasta-headers.mdx
+++ b/lapis-docs/src/content/docs/concepts/customizable-fasta-headers.mdx
@@ -1,0 +1,55 @@
+---
+title: Customizable FASTA headers
+description: Customize the header line of FASTA records returned by sequence endpoints
+---
+
+When requesting sequences as FASTA, you can customize the header line of each record by setting the
+`fastaHeaderTemplate` request property.
+The parameter is ignored if the data format is not FASTA.
+
+The template allows placeholders in curly braces:
+
+- `{<metadataField>}` — replaced with the value of the metadata field for that sequence (e.g. `{country}`).
+  The instance's primary key field (configured by the maintainer) can be used as a placeholder in the same way.
+- `{.segment}` — replaced with the segment name. Only valid on nucleotide sequence endpoints.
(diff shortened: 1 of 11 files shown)


lapis-docs/src/content/docs/concepts/customizable-fasta-headers.mdx
```diff
@@ -0,0 +1,55 @@
+---
+title: Customizable FASTA headers
+description: Customize the header line of FASTA records returned by sequence endpoints
+---
+
+When requesting sequences as FASTA, you can customize the header line of each record by setting the
+`fastaHeaderTemplate` request property.
+The parameter is ignored if the data format is not FASTA.
```
fengelniederhammer: LAPIS returns an error in this case.

lapis-docs/src/content/docs/concepts/lineage-queries.mdx
```diff
@@ -0,0 +1,48 @@
+---
+title: Lineage queries
+description: Filtering lineage fields and their descendants
+---
+
+import { OnlyIf } from '../../../components/OnlyIf.tsx';
+import { getConfig, hasLineageFields } from '../../../config.ts';
+import { Code } from '@astrojs/starlight/components';
+
+export const lineageField = 'lineageField';
```
fengelniederhammer, line 10: We usually tried to write the actual field names into the example, since we assume that the docs page runs next to an actual LAPIS instance. We could search the config for a lineage field and only use this hardcoded value as fallback

chaoran-chen: I have been thinking about this and I'm currently a bit tending towards the direction of minimizing instance-specific docs and trying more to write the docs in a general way. This keeps the docs simpler to maintain / avoid bugs and most of the currently running LAPISes don't deploy the documentation anyways (since Loculus doesn't deploy them).

lapis-docs/src/content/docs/concepts/customizable-fasta-headers.mdx
chaoran-chen:
```suggestion
An error is thrown if the prameter is set but the data format is not FASTA.
```

lapis-docs/src/content/docs/concepts/lineage-queries.mdx
fengelniederhammer: Ok, let's discuss this tomorrow then.

lapis-docs/src/content/docs/concepts/customizable-fasta-headers.mdx
fengelniederhammer, line 1: Maybe an example on how the result would look like might also help? (Non blocking, we can also do that in a separate PR)

fengelniederhammer (approved):

lapis-docs/src/content/docs/concepts/customizable-fasta-headers.mdx
chaoran-chen: (Doing this potentially in a future PR and merging this to not keep it here indefinitely)

PR #1760, opened 2026-07-07
refactor(lapis): add abstract SaneQL serialization layer

### LINKED ISSUE #1754: refactor(lapis): SaneQL object layer
## Now

`SiloQuery.kt` = objects for old (pre-SaneQL) SILO query. Trivial JSON.

`SiloQuerySaneQlSerializer`: SiloQuery objects -> SaneQL string in one big
`when`. Builds strings inline. Mixes 2 jobs: pick SaneQL + write SaneQL
text (quoting, escaping, precedence).

## Want

New middle layer: SaneQL objects (small AST). Trivial render to string.

Two steps:
1. SiloQuery -> SaneQL AST  (map)
2. SaneQL AST -> string     (render, trivial)

## Why

- Split "what SaneQL" from "how text look".
- Quoting/escaping/precedence in one place, not scattered.
- AST testable alone.

## Scope

Pure refactor. Output byte-identical. `SiloQuerySaneQlSerializerTest` stays
green. `SiloClient` entry point unchanged.

## AST sketch

- Exprs: literals, identifiers, binary (`=`, `&&`, `||`), unary (`!`),
  func calls (`nucleotideEquals(..)`, `isNull(..)`, `nOf(..)`), method calls
  (`col.between(..)`, `col.desc()`), list literals (`{a, b}`).
- Pipeline: `default.filter(expr)` + steps (`.groupBy`, `.map`, `.mutations`,
  `.project`, `.orderBy`, `.offset`, `.randomize`, `.limit`).
- render() owns: id quoting `"`->`""`, string `'`->`''`, date `::date`,
  And/Or parens (today's `wrapIfNeeded`).

fengelniederhammer:
resolves #1754

Adds a reusable representation of SaneQL syntax in the code as an intermediate layer between the SILO expression object tree and the SaneQL string.

## PR Checklist


lapis/src/main/kotlin/org/genspectrum/lapis/silo/SaneQlAst.kt
```diff
@@ -0,0 +1,162 @@
...
+    val namedArgs: List<SaneQlNamedArg> = emptyList(),
+) : SaneQlExpression {
+    override fun render() = "${receiver.render()}.$name(${renderArgs(positionalArgs, namedArgs)})"
+}
+
+private fun renderArgs(
+    positionalArgs: List<SaneQlExpression>,
+    namedArgs: List<SaneQlNamedArg>,
+): String =
+    (positionalArgs.map { it.render() } + namedArgs.map { "${it.name}:=${it.value.render()}" })
```
fhennig: Why don't the named args above also have a render function? I think it'd be more consistent if they also could render themselves.

fhennig (approved):

PR #1768, opened 2026-07-14
feat(lapis): add sequence position fields syntax to aggregated endpoint

### LINKED ISSUE #1742: Query co-occurences from SILO
SILO now supports these queries:

```
default.map({pos_1 := sequence.at(1), ..., pos_421 := sequence.at(421)}).groupBy({count := count()}, {pos_1, ..., pos_421})
```

which will return a table with all present combinations of symbols at the listed positions.

LAPIS should expose this functionality in a specialized endpoint, which receives a sequence name and a list of positions from the user and queries SILO with exactly the query listed above

### LINKED ISSUE #1766: Add aggregation by sequence positions
As part of #1765 we should enable the aggregation by sequence positions. The syntax should be `S[123]` for genes. `main` should be able to be omitted

fhennig:
## Summary

- Adds `SequenceName[position]` syntax (e.g. `S[501]`) to the `fields` parameter of `/aggregated`, allowing users to request specific sequence positions as group-by columns
- Shorthand `[N]` is supported for single-segmented genomes and resolves to the single nucleotide sequence
- Sequence names are validated case-insensitively against both genes and nucleotide segments from the reference genome
- `/details`, phylo tree, and MRCA endpoints reject position fields at the type level (they only ever construct a `PlainField`), with a clear error message
- GET `/aggregated`'s `fields` query parameter now documents the position-field syntax too, matching what the POST body schema already described
- Generates a `.map({"S[501]":="S".at(501)})` SaneQL step before `.groupBy(...)` when position fields are present — the user-facing name is used directly as the SaneQL alias, so no post-processing is needed
- Response column keys mirror the input exactly: `S[501]` in → `S[501]` out, `[501]` in → `[501]` out
- Internal cleanup: split the field/orderBy converters and cleaners out of `Field.kt`/`OrderByField.kt` into a dedicated `request.converter` package, renamed the sealed field hierarchy for clarity (`RequestField` → `Field`, `Field` → `PlainField`), and renamed `SequenceFiltersRequestWithFields` → `AggregatedFiltersRequest` to match its `/aggregated`-only sibling naming (`DetailsFiltersRequest`, `PhyloTreeSequenceFiltersRequest`, `MRCASequenceFiltersRequest`)

## Example

```
POST /sample/aggregated
{"fields": ["country", "S[501]"]}
```

Generates:
```
default.filter(true).map({"S[501]":="S".at(501)}).groupBy({"count":=count()}, {"country", "S[501]"})
```

## Test plan

- [x] Unit tests for `SequenceName[position]` parsing in `AggregatedFiltersRequestTest`
- [x] Error cases: unknown sequence name, position ≤ 0, shorthand on multi-segmented genome
- [x] SaneQL generation tests in `SiloQueryToSaneQlTest` for position-only and mixed fields
- [x] e2e coverage for a gene (amino acid) position, a plain nucleotide position, and the single-segment shorthand syntax
- [x] Full test suite passes


fhennig:
~~Ok, something weird with the test? But still reviewable I think. I'll look into it.~~

lapis/src/main/kotlin/org/genspectrum/lapis/controller/ControllerDescriptions.kt
```diff
@@ -68,7 +68,11 @@ such as combinations of mutations ("mutation 1 or mutation 2") and other filters
 const val AGGREGATED_GROUP_BY_FIELDS_DESCRIPTION =
     """The fields to stratify by.
 If empty, only the overall count is returned.
-If requesting CSV or TSV data, the columns are ordered in the same order as the fields are specified here."""
+If requesting CSV or TSV data, the columns are ordered in the same order as the fields are specified here.
+
+Sequence positions can be requested using the syntax `SequenceName[position]` (e.g. `S[501]` for position 501 of sequence `S`).
+For single-segmented genomes, the shorthand `[position]` (e.g. `[501]`) can be used.
+Position field column names in the response use the canonical sequence name from the reference genome (case-insensitive input), e.g. `s[501]` -> `S[501]`."""
```
fengelniederhammer: We should probably also add something in the llms.txt?

fengelniederhammer: Usually we tried to avoid those formulations (if single segmented ...). Instead, we generate them at run time: then we know in which case we are.

lapis/src/main/kotlin/org/genspectrum/lapis/request/Field.kt
```diff
@@ -17,25 +37,61 @@ fun interface FieldConverter<T> {
...
         val cleaned = caseInsensitiveFieldsCleaner.clean(source)
             ?: throw BadRequestException(
                 "Unknown field: '$source', known values are ${caseInsensitiveFieldsCleaner.getKnownFields()}",
             )
-
         return Field(cleaned)
     }
 
-    override fun validatePhyloTreeFields(source: String): Field {
+    override fun validatePhyloTreeFields(source: String): RequestField {
```
fengelniederhammer: Isn't this function actually duplicate? Below, we have another `validatePhyloTreeField` that looks like basically the same.

fengelniederhammer: I think this one isn't even used anywhere...

lapis/src/main/kotlin/org/genspectrum/lapis/request/Field.kt
```diff
@@ -1,12 +1,32 @@
 package org.genspectrum.lapis.request
 
 import org.genspectrum.lapis.config.DatabaseConfig
+import org.genspectrum.lapis.config.ReferenceGenomeSchema
 import org.genspectrum.lapis.controller.BadRequestException
 import org.springframework.stereotype.Component
 
+private val SEQUENCE_POSITION_REGEX = Regex("""^([A-Za-z][A-Za-z0-9_]*)\[(\d+)\]$""")
+private val SHORTHAND_POSITION_REGEX = Regex("""^\[(\d+)\]$""")
```
fengelniederhammer:
```suggestion
private val SEQUENCE_POSITION_REGEX = Regex("""^([A-Za-z][A-Za-z0-9_]*)\[(\d+)]$""")
private val SHORTHAND_POSITION_REGEX = Regex("""^\[(\d+)]$""")
```
apparently those closing brackets don't need to be escaped.

lapis/src/main/kotlin/org/genspectrum/lapis/request/Field.kt
```diff
@@ -1,12 +1,32 @@
...
+
+data class SequencePositionField(
+    val sequenceName: String,
+    val position: Int,
+    val isSingleSegment: Boolean = false,
+) : RequestField {
+    /** Name used both as the SaneQL alias and as the response column key, e.g. `S[501]` or `[501]` for shorthand. */
+    val userFacingName: String get() = if (isSingleSegment) "[$position]" else "$sequenceName[$position]"
+    override val outputColumnName: String get() = userFacingName
+}
```
fengelniederhammer, line 20:
```suggestion
data class SequencePositionField(
    val sequenceName: String?,
    val position: Int,
) : RequestField {
    override val outputColumnName: String = if (sequenceName == null) {
        "[$position]"
    } else {
        "$sequenceName[$position]"
    }
}
```

We can encode `isSingleSegment` into `sequenceName` by making it nullable.

lapis/src/main/kotlin/org/genspectrum/lapis/request/Field.kt
```diff
@@ -17,25 +37,61 @@ fun interface FieldConverter<T> {
...
+                throw BadRequestException("Invalid position in '$source': must be a positive integer, got $position")
+            }
+            val canonicalName = referenceGenomeSchema.getSequenceNameFromCaseInsensitiveName(name)
+                ?: throw BadRequestException(
+                    "Unknown sequence '$name' in '$source', known sequences are: " +
+                        (referenceGenomeSchema.getNucleotideSequenceNames() + referenceGenomeSchema.getGeneNames())
+                            .joinToString(", "),
+                )
+            return SequencePositionField(canonicalName, position)
+        }
```
fengelniederhammer: Could we collapse those two branches by making the sequence name in the regex optional? The position parsing is the same, and we could streamline the sequence name parsing.

lapis/src/main/kotlin/org/genspectrum/lapis/controller/LapisController.kt
```diff
@@ -1490,7 +1492,14 @@ class LapisController(
     }
 
     private fun getDetailsCollection(request: SequenceFiltersRequestWithFields): DetailsCollection {
-        val fields = request.fields.map { it.fieldName }
+        val positionFields = request.fields.filterIsInstance<SequencePositionField>()
+        if (positionFields.isNotEmpty()) {
+            throw BadRequestException(
+                "Sequence position fields are not supported for this endpoint: " +
+                    positionFields.joinToString(", ") { it.userFacingName },
+            )
```
fengelniederhammer: I find it a bit weird to first allow a `SequencePositionField` while parsing and then reject it. Wouldn't it be better to differentiate already in the types that the controller methods accept?

lapis/src/main/kotlin/org/genspectrum/lapis/silo/SiloQuery.kt
```diff
@@ -235,17 +239,42 @@ sealed class SiloAction<ResponseType>(
...
+                                SaneQlList(
+                                    sequencePositionFields.map { field ->
+                                        SaneQlAssignment(
+                                            field.userFacingName,
+                                            SaneQlMethodCall(
+                                                SaneQlIdentifier(field.sequenceName),
+                                                "at",
+                                                listOf(SaneQlInt(field.position)),
+                                            ),
+                                        )
```
fengelniederhammer, line 257: Do we need to deduplicate those somewhere?

fengelniederhammer (changes requested):
I'm also missing an e2e test ;)

lapis/src/main/kotlin/org/genspectrum/lapis/request/Field.kt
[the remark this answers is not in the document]
fengelniederhammer: No, I think this is unnecessary.

lapis/src/main/kotlin/org/genspectrum/lapis/request/Field.kt
fhennig: For downstream construction of the SILO query, it's better to have the resolved 'main' sequence name in here already - otherwise we have to do the resolution there, where we don't currently use the ReferenceGenomeSchema.

It's not overly verbose or anything, I'd say let's keep it, to not scatter the resolution mechanism across multiple places.

lapis/src/main/kotlin/org/genspectrum/lapis/silo/SiloQuery.kt
fhennig: done.

lapis/src/main/kotlin/org/genspectrum/lapis/request/DetailsFiltersRequest.kt
```diff
@@ -0,0 +1,64 @@
...
+            parsedCommonFields.sequenceFilters,
+            parsedCommonFields.nucleotideMutations,
+            parsedCommonFields.aminoAcidMutations,
+            parsedCommonFields.nucleotideInsertions,
+            parsedCommonFields.aminoAcidInsertions,
+            fields,
+            parsedCommonFields.orderByFields,
+            parsedCommonFields.limit,
+            parsedCommonFields.offset,
+        )
```
fengelniederhammer, line 45: Let's use named arguments in this constructor call?

lapis/src/main/kotlin/org/genspectrum/lapis/request/DetailsFiltersRequest.kt
```diff
@@ -0,0 +1,64 @@
...
+    private val caseInsensitiveFieldConverter: CaseInsensitiveFieldConverter,
+) : FieldConverter<Field> {
+    override fun convert(source: String): Field {
+        val converted = caseInsensitiveFieldConverter.convert(source)
+        if (converted !is Field) {
+            throw BadRequestException(
+                "Sequence position fields are not supported for this endpoint: ${converted.outputColumnName}",
+            )
+        }
+        return converted
```
fengelniederhammer: As discussed: Let's get rid of this check here by making the aggregated parsing the special case on top.

lapis/src/main/kotlin/org/genspectrum/lapis/request/AggregatedFiltersRequest.kt
fengelniederhammer, line 1: Maybe we should also make this explicit and rename it to `AggregatedRequest` or something?

lapis/src/main/kotlin/org/genspectrum/lapis/request/Field.kt
```diff
@@ -1,53 +1,89 @@
...
+}
+
+data class SequencePositionField(
+    val sequenceName: String,
+    val position: Int,
+    val isSingleSegment: Boolean = false,
+) : RequestField {
+    /** Name used both as the SaneQL alias and as the response column key, e.g. `S[501]` or `[501]` for shorthand. */
+    val userFacingName: String get() = if (isSingleSegment) "[$position]" else "$sequenceName[$position]"
+    override val outputColumnName: String get() = userFacingName
```
fengelniederhammer: Do we still need the distinction between `userFacingName` and `outputColumnName`? We can probably get rid of `userFacingName` (and inline it into `outputColumnName`)?

lapis/src/test/kotlin/org/genspectrum/lapis/controller/LandingPageControllerTest.kt
```diff
@@ -45,4 +45,12 @@ class LandingPageControllerTest(
...
             .andExpect(content().string(not(containsString("$")))) // dollars hint at broken Thymeleaf template
     }
+
+    @Test
+    fun `WHEN calling llms txt endpoint THEN documents sequence position fields`() {
+        mockMvc.perform(get("/llms.txt"))
+            .andExpect(status().isOk)
+            .andExpect(content().string(containsString("Sequence Position Fields")))
+            .andExpect(content().string(containsString("SequenceName[position]")))
+    }
```
fengelniederhammer: (Nit: Doesn't hurt, but IMO we don't necessarily need this test)

lapis/src/main/kotlin/org/genspectrum/lapis/openApi/OpenApiDocs.kt
```diff
@@ -386,6 +398,19 @@ fun buildOpenApiSchema(
                 .addSchemas(FORMAT_SCHEMA, dataFormatSchema()),
         )
 
+internal fun aggregatedFieldsDescription(referenceGenomeSchema: ReferenceGenomeSchema): String {
```
fengelniederhammer, line 401: The GET endpoint still has the old description: 

Image

Is it possible to use this there, too?

lapis-e2e/test/aggregatedQueries/sequencePositionField.json
fengelniederhammer, line 1: Let's also add a test that groups by a nuc position?

lapis/src/main/kotlin/org/genspectrum/lapis/config/DatabaseConfigValidator.kt
```diff
@@ -11,6 +11,11 @@ class DatabaseConfigValidator {
                     "Metadata field name '${it.name}' contains the reserved character '.'",
                 )
             }
+            if (it.name.contains('[') || it.name.contains(']')) {
+                throw IllegalArgumentException(
+                    "Metadata field name '${it.name}' contains a reserved character, '[' or ']'",
+                )
```
fengelniederhammer, line 17: 👍

lapis/src/main/kotlin/org/genspectrum/lapis/LapisSpringConfig.kt
```diff
@@ -115,6 +118,21 @@ class LapisSpringConfig {
             }
         }
 
+    /**
+     * The GET `fields` parameter is documented via the static `@FieldsToAggregateBy` annotation, which can't hold
+     * the dynamic [aggregatedFieldsDescription] (it depends on the reference genome). Overwrite it here after the
+     * fact, matching what the POST body's `AGGREGATED_REQUEST_SCHEMA` already documents.
```
fengelniederhammer: Then let's maybe change the original description in the annotation in the controller? Either remove it or make it say "I'm just a placeholder that will be overwritten later"?

lapis/src/main/kotlin/org/genspectrum/lapis/LapisSpringConfig.kt
```diff
@@ -115,6 +118,21 @@ class LapisSpringConfig {
...
+     */
+    @Bean
+    fun aggregatedFieldsGetCustomizer(referenceGenomeSchema: ReferenceGenomeSchema) =
+        OpenApiCustomizer { openApi ->
+            openApi.paths["/sample$AGGREGATED_ROUTE"]
+                ?.get
+                ?.parameters
+                ?.find { it.name == FIELDS_PROPERTY }
+                ?.description = aggregatedFieldsDescription(referenceGenomeSchema)
```
fengelniederhammer: I just noticed that we have a `SampleEndpointsGetParameterCustomizer`, maybe we should put it there?

Also the SwaggerUI of the multi segmented local instance doesn't load. Generating the API docs fails:
```
2026-07-28 08:53:57,290 WARN [http-nio-8094-exec-4] [4bcc64a3-e491-4a89-8278-d33272b0fb65] org.genspectrum.lapis.controller.ExceptionHandler: Caught unexpected exception: getContent(...) must not be null
java.lang.NullPointerException: getContent(...) must not be null
        at org.genspectrum.lapis.openApi.OperationsSorter.customize$lambda$0(OperationsSorter.kt:18)
        at org.genspectrum.lapis.openApi.OperationsSorter.customize$lambda$1(OperationsSorter.kt:16)
        at java.base/java.util.LinkedHashMap.forEach(LinkedHashMap.java:986)
        at org.genspectrum.lapis.openApi.OperationsSorter.customize(OperationsSorter.kt:16)
```

fhennig: I pushed a fix.

fhennig:
Failures are also on main. "CI pulled a newer SILO release that changed its API contract for single-segmented genomes
  — it now requires an explicit sequenceName argument for mutation queries, and includes the main: segment prefix in insertion output, where it used to default/omit it."

lapis/src/main/kotlin/org/genspectrum/lapis/openApi/OpenApiDocs.kt
```diff
@@ -386,6 +398,19 @@ fun buildOpenApiSchema(
...
+    val shorthand = if (referenceGenomeSchema.isSingleSegmented()) {
+        " The shorthand `[position]` (e.g. `[501]`) can also be used."
```
fengelniederhammer:
```suggestion
        " The shorthand `[position]` (e.g. `[501]`) can also be used for nucleotide sequences."
```

lapis/src/main/kotlin/org/genspectrum/lapis/openApi/OpenApiDocs.kt
```diff
@@ -386,6 +398,19 @@ fun buildOpenApiSchema(
...
+    } else {
+        ""
+    }
+    return AGGREGATED_GROUP_BY_FIELDS_DESCRIPTION +
```
fengelniederhammer, line 407: Nit: We could inline `AGGREGATED_GROUP_BY_FIELDS_DESCRIPTION` now.

lapis/src/main/kotlin/org/genspectrum/lapis/openApi/OpenApiDocs.kt
```diff
@@ -386,6 +398,19 @@ fun buildOpenApiSchema(
...
+        "\n\nSequence positions can be requested using the syntax `SequenceName[position]` " +
+        "(e.g. `S[501]` for position 501 of sequence `S`).$shorthand\n" +
+        "Position field column names in the response use the canonical sequence name from the reference genome " +
+        "(case-insensitive input), e.g. `s[501]` -> `S[501]`."
```
fengelniederhammer, line 411: I think I would add that you can use any metadata field here (previously that was clear from the enum type, now it's not and the description only mentions sequence positions)

fengelniederhammer (approved):

PR #1780, opened 2026-07-23
feat(lapis): support scalar functions in aggregation fields via dot notation

### LINKED ISSUE #1767: Add scalar function to aggregation fields
`aggregation` fields should be able to contain SILO scalar functions.

e.g. `date.isoWeek` should map to `date.isoWeek()` in SILO. This should then be used as an aggregation field

fhennig:
resolves #1767 

### Summary

Users can now specify computed fields like `date.isoWeek` in the `fields` parameter of the `/aggregated` endpoint. LAPIS validates the function against a whitelist (currently `isoWeek` for date fields), generates a `map()` step before `groupBy()` in the SaneQL query, and renames the internal alias columns back to the user-facing `field.function` names in the response.

## PR Checklist

### DIFF
--- a/lapis-docs/src/content/docs/concepts/computed-fields.mdx
+++ b/lapis-docs/src/content/docs/concepts/computed-fields.mdx
@@ -31,7 +31,6 @@ The response uses the full `<field>.<function>` string as the column/property ke
 | `isoWeek` | `date`                 | The ISO 8601 week of the date field. |
 
 :::note
-Computed fields are only supported in the `fields` parameter of `/sample/aggregated`
-(and the analogous endpoints for other sequence types).
+Computed fields are only supported in the `fields` parameter of `/sample/aggregated`.
 They are not supported by endpoints that return the underlying data rows, such as `/sample/details`.
 :::
--- a/lapis-docs/src/content/docs/concepts/computed-fields.mdx
+++ b/lapis-docs/src/content/docs/concepts/computed-fields.mdx
@@ -25,8 +25,8 @@ The response uses the full `<field>.<function>` string as the column/property ke
 ## Available functions
 
 | Function  | Applicable field types | Description                          |
-|-----------|-------------------------|---------------------------------------|
-| `isoWeek` | `date`                  | The ISO 8601 week of the date field. |
+| --------- | ---------------------- | ------------------------------------ |
+| `isoWeek` | `date`                 | The ISO 8601 week of the date field. |
 
 :::note
 Computed fields are only supported in the `fields` parameter of `/sample/aggregated`
--- a/lapis-docs/src/content/docs/concepts/computed-fields.mdx
+++ b/lapis-docs/src/content/docs/concepts/computed-fields.mdx
@@ -10,11 +10,13 @@ instead of grouping by the raw field value.
 
 **Example:** to count sequences per ISO week instead of per exact date:
 
-```
-[URL to LAPIS instance]/sample/aggregated?fields=date.isoWeek
+```http
+GET /sample/aggregated?fields=date.isoWeek
 ```
 
-```json
+```http
+POST /sample/aggregated
+
 {
     "fields": ["date.isoWeek"]
 }
--- a/lapis-docs/src/content/docs/concepts/computed-fields.mdx
+++ b/lapis-docs/src/content/docs/concepts/computed-fields.mdx
@@ -1,0 +1,35 @@
+---
+title: Computed fields
+description: Computed fields
+---
+
+In addition to plain metadata field names, the `fields` parameter of `/sample/aggregated` accepts computed fields
+using the syntax `<field>.<function>`, e.g. `date.isoWeek`.
+A computed field applies a scalar function to a metadata field and groups by the result,
+instead of grouping by the raw field value.
+
+**Example:** to count sequences per ISO week instead of per exact date:
+
+```
+[URL to LAPIS instance]/sample/aggregated?fields=date.isoWeek
+```
+
+```json
+{
+    "fields": ["date.isoWeek"]
+}
+```
+
+The response uses the full `<field>.<function>` string as the column/property key, e.g. `date.isoWeek`.
+
+## Available functions
+
+| Function  | Applicable field types | Description                          |
+|-----------|-------------------------|---------------------------------------|
+| `isoWeek` | `date`                  | The ISO 8601 week of the date field. |
+
+:::note
+Computed fields are only supported in the `fields` parameter of `/sample/aggregated`
+(and the analogous endpoints for other sequence types).
+They are not supported by endpoints that return the underlying data rows, such as `/sample/details`.
+:::
--- a/lapis-e2e/test/aggregated.spec.ts
+++ b/lapis-e2e/test/aggregated.spec.ts
@@ -172,6 +172,56 @@ describe('The /aggregated endpoint', () => {
     expect(resultJson.error.detail).to.include("Unknown field: 'notAField', known values are [primaryKey,");
   });
 
+  it('should stratify by a computed field using dot notation', async () => {
+    const result = await lapisClient.postAggregated({
+      aggregatedPostRequest: {
+        date: '2021-06-05',
+        fields: ['date.isoWeek'],
+      },
+    });
+
+    expect(result.data).to.have.length(1);
+    expect(result.data[0]).to.have.property('count', 1);
+    expect(result.data[0]).to.have.property('date.isoWeek', 22);
+  });
+
+  it('should order by a computed field using dot notation', async () => {
+    const result = await lapisClient.postAggregated({
+      aggregatedPostRequest: {
+        date: '2021-06-05',
+        fields: ['date.isoWeek'],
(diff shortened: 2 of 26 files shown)


fhennig:
## Code review (high effort)

Reviewed `git diff main...HEAD` for this branch. One confirmed correctness bug, plus a few lower-severity cleanup findings.

### 🐞 Correctness bug (CONFIRMED)

**`orderBy` references to computed fields are not case-normalized the way `fields` entries are** — `lapis/src/main/kotlin/org/genspectrum/lapis/model/SiloQueryModel.kt:74`

`OrderByFieldsCleaner.clean` (`request/OrderByField.kt:148`) only delegates to `CaseInsensitiveFieldsCleaner.clean`, which has no awareness of the `<field>.<function>` dot syntax and no dot-splitting logic. For a dotted orderBy value it returns `null` and falls back (`?: fieldName`) to the raw, un-normalized string the client sent — unlike `fields`, which always goes through `CaseInsensitiveFieldConverter.convert` and produces a canonical `"$sourceField.${function.saneQlMethodName}"` string.

`rewriteOrderByForComputedFields`'s `userFacingToAlias[field.field]` lookup is an exact, case-sensitive match against that canonical form.

**Repro:** `fields=["date.isoWeek"]`, `orderBy=[{field:"DATE.ISOWEEK"}]` (or any casing variant differing from the canonical form). The alias lookup misses, the substitution silently doesn't happen, and the raw string `"DATE.ISOWEEK"` is emitted verbatim via `id()` as a literal SaneQL column identifier — a column that doesn't exist in the post-`groupBy` result (only `date`/`__scalar_isoWeek_date` exist), causing an unexpected SILO-side error instead of the case-insensitive behavior the rest of the API guarantees for plain fields.

None of the added tests exercise this path — `SiloQueryModelTest` constructs the orderBy field directly from `isoWeekField.fieldName` in memory, bypassing the real cleaner/converter entirely, so the mismatch is untested.

### 🧹 Cleanup findings (lower severity)

1. **Repeated scans of the same list** — `SiloQueryModel.kt:43-55`. `getAggregated` calls `sequenceFilters.fields.filterIsInstance<Field.Plain>()` / `filterIsInstance<Field.Computed>()` four separate times over the same list instead of partitioning once. Minor per-request cost, and any future change to how computed fields are derived needs updating in three call sites instead of one.

2. **Duplicated lowercase lookup map** — `request/Field.kt:38-39`. `fieldTypesByLowercaseName` rebuilds the same `metadata.associateBy(name.lowercase())` that `CaseInsensitiveFieldsCleaner.fieldsMap` already builds from the identical source list. Currently safe since both derive from the same source, but the non-null assertion at `Field.kt:69` relies on that invariant holding — if the two maps ever drift, a `BadRequestException` turns into an unhandled NPE.

3. **Computed-field support is bolted onto `AggregatedAction` only** — `silo/SiloQuery.kt:229` onward. The `computedFields`/map-step/alias/rename/orderBy-rewrite logic is wired specifically into `AggregatedAction` and `SiloQueryModel`, with other endpoints rejecting computed fields via an ad hoc throw in the controller (`LapisController.kt:1494-1499`) rather than a shared capability. If scalar functions are later extended to another endpoint, this whole mechanism will likely be copy-pasted rather than reused.

---
Note: I also checked whether the dot-notation parsing regresses metadata fields whose names literally contain a `.` (since `convert()` treats any dotted string as computed-field syntax) — this is **not an issue**, `DatabaseConfigValidator` already rejects `.` in field names at config-load time, predating this branch.

lapis-docs/src/content/docs/concepts/computed-fields.mdx
```diff
@@ -0,0 +1,35 @@
...
+
+## Available functions
+
+| Function  | Applicable field types | Description                          |
+| --------- | ---------------------- | ------------------------------------ |
+| `isoWeek` | `date`                 | The ISO 8601 week of the date field. |
+
+:::note
+Computed fields are only supported in the `fields` parameter of `/sample/aggregated`
+(and the analogous endpoints for other sequence types).
```
fengelniederhammer, line 33: which other endpoint?

lapis-docs/src/content/docs/concepts/computed-fields.mdx
```diff
@@ -0,0 +1,35 @@
...
+}
+```
+
+The response uses the full `<field>.<function>` string as the column/property key, e.g. `date.isoWeek`.
+
+## Available functions
+
+| Function  | Applicable field types | Description                          |
+| --------- | ---------------------- | ------------------------------------ |
+| `isoWeek` | `date`                 | The ISO 8601 week of the date field. |
```
fengelniederhammer, line 29: Image

Something messed up the table formatting...

lapis-docs/src/content/docs/concepts/computed-fields.mdx
```diff
@@ -0,0 +1,35 @@
...
+
+```
+[URL to LAPIS instance]/sample/aggregated?fields=date.isoWeek
+```
+
+```json
+{
+    "fields": ["date.isoWeek"]
+}
+```
```
fengelniederhammer, line 21:
````suggestion
```
POST /sample/aggregated?fields=date.isoWeek

{
    "fields": ["date.isoWeek"]
}
```
````

I think other examples also already use a format like this?

File: lapis-e2e/test/aggregated.spec.ts [hunk 4]
```diff
@@ -172,6 +172,56 @@ describe('The /aggregated endpoint', () => {
...
+
+    expect(result.data).to.have.length(1);
+    expect(result.data[0]).to.have.property('count', 1);
+    expect(result.data[0]).to.have.property('date.isoWeek', 22);
+  });
+
+  it('should order by a computed field using dot notation', async () => {
+    const result = await lapisClient.postAggregated({
+      aggregatedPostRequest: {
+        date: '2021-06-05',
```
fengelniederhammer, line 191: really filter by date, too? I think the test would be more meaningful without the filter, but still deterministic due to the orderBy.

lapis/src/main/kotlin/org/genspectrum/lapis/request/converter/ScalarFunctionFieldConverter.kt
```diff
@@ -0,0 +1,46 @@
...
+import org.genspectrum.lapis.config.MetadataType
+import org.genspectrum.lapis.controller.BadRequestException
+import org.genspectrum.lapis.request.ComputedField
+import org.genspectrum.lapis.request.ScalarFunction
+import org.springframework.stereotype.Component
+
+@Component
+class ScalarFunctionFieldConverter(
+    private val caseInsensitiveFieldsCleaner: CaseInsensitiveFieldsCleaner,
+    private val databaseConfig: DatabaseConfig,
```
fengelniederhammer, line 13:
```suggestion
    private databaseConfig: DatabaseConfig,
```

lapis/src/test/kotlin/org/genspectrum/lapis/request/ScalarFunctionFieldTest.kt
```diff
@@ -0,0 +1,69 @@
...
+
+    @Test
+    fun `unknown function name throws BadRequestException`() {
+        assertThrows<BadRequestException> { underTest.tryConvert("date.unknownFunction") }
+    }
+
+    @Test
+    fun `isoWeek on non-date field throws BadRequestException`() {
+        val ex = assertThrows<BadRequestException> { underTest.tryConvert("country.isoWeek") }
+        assertTrue(ex.message.orEmpty().contains("STRING"), "Expected error to mention type, got: ${ex.message}")
```
fengelniederhammer, line 67:
```suggestion
        assertTrue(ex.message.contains("STRING"), "Expected error to mention type, got: ${ex.message}")
```

According to my IDE, this is redundant.

lapis/src/test/kotlin/org/genspectrum/lapis/request/ScalarFunctionFieldTest.kt
```diff
@@ -0,0 +1,69 @@
...
+private val testDatabaseConfig = databaseConfig(
+    primaryKey = "accession",
+    metadata = listOf(
+        DatabaseMetadata("accession", MetadataType.STRING),
+        DatabaseMetadata("date", MetadataType.DATE),
+        DatabaseMetadata("country", MetadataType.STRING),
+    ),
+)
+
+private val underTest = ScalarFunctionFieldConverter(
```
fengelniederhammer, line 24: Nit: Usually, in Kotlin and Java, those `underTest`s  are declared inside the test class.

fengelniederhammer (approved):
A couple of minor things, looks good otherwise 👍

lapis-docs/src/content/docs/concepts/computed-fields.mdx
fhennig: I don't see this locally.

taepper: I cannot see this error either

PR #1798, opened 2026-07-29
fix(lapis): restore `insertion` and `mutation` format in response

### LINKED ISSUE #1797: restore `insertion` and `mutation` format in response
https://github.com/GenSpectrum/LAPIS-SILO/issues/1403 removes the `insertion` field from insertion responses and `mutation` from mutations responses. Restore the behavior in LAPIS (we can simply assemble the field ourselves) to how it was before https://github.com/GenSpectrum/LAPIS-SILO/pull/1386:
* insertions: `ins_<seq_name>:<position>:<symbols>` (`ins_<position>:<symbols>` when single segmented)
* mutations: `<seq_name>:<from><positions><to>` (`<from><positions><to>` when single segmented)

fengelniederhammer:
resolves #1797

### Summary

SILO dropped the assembled `insertion` and `mutation` fields from its responses. This restores them by assembling them in LAPIS, keeping the API contract unchanged.

SILO also dropped the concept of a "default" sequence, making the `sequenceName' field in insertion and mutation filters mandatory.

## PR Checklist

### DIFF
--- a/lapis/src/main/kotlin/org/genspectrum/lapis/request/NucleotideMutation.kt
+++ b/lapis/src/main/kotlin/org/genspectrum/lapis/request/NucleotideMutation.kt
@@ -43,6 +43,13 @@ data class NucleotideMutation(
                 }
                 ?.name
 
+            if (segmentName == null && !referenceGenomeSchema.isSingleSegmented()) {
+                throw BadRequestException(
+                    "The reference genome is multi-segmented, but no segment was specified in the nucleotide " +
+                        "mutation '$nucleotideMutation'. Please specify one, e.g. 'segmentName:A123T'.",
+                )
+            }
+
             return NucleotideMutation(
                 segmentName,
                 position,
--- a/.github/workflows/lapis.yml
+++ b/.github/workflows/lapis.yml
@@ -4,10 +4,7 @@ on: [ push ]
 
 env:
   DOCKER_IMAGE_NAME: ghcr.io/genspectrum/lapis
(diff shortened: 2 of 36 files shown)


lapis/src/main/kotlin/org/genspectrum/lapis/request/NucleotideMutation.kt
```diff
@@ -43,6 +43,13 @@ data class NucleotideMutation(
                 }
                 ?.name
 
+            if (segmentName == null && !referenceGenomeSchema.isSingleSegmented()) {
+                throw BadRequestException(
+                    "The reference genome is multi-segmented, but no segment was specified in the nucleotide " +
+                        "mutation '$nucleotideMutation'. Please specify one, e.g. 'segmentName:A123T'.",
+                )
+            }
```
taepper, line 51: Ah, before this could have succeeded because SILO could have a default? Good that we cleaned this up to LAPIS's single definition re/ omission of sequence names

taepper (approved):
LGTM! AFAICT most changes regard the removal of default sequence name in SILO, maybe also add this to the PR description?

lapis/src/main/kotlin/org/genspectrum/lapis/request/NucleotideMutation.kt
fengelniederhammer: Not sure, I think SILO would have thrown an error (there are many instances where we rely on SILO validation rather than validating ourselves in LAPIS again).

PR #1850, opened 2026-09-03
fix(lapis): do not cache aggregations with many group by fields

### LINKED ISSUE #1849: Disable caching for large/overly-specific queries
We ran into a problem with queries that request a lot of positions to group by. The query itself is very specific and has a large response. Since it's an `AggregatedAction` it is set as cacheable. This also means it is materialized (not streamed) and then cached - it takes up space to materialize and it takes up space in the cache.

In `SiloQuery.kt` in `AggregatedAction` we set `cacheable = true` unconditionally. This means that requests that us a lot of group by fields are also cached, even if we're grouping by ~400 position fields.

It's unlikely that we will get cache hits of such specific queries, so they just end up taking up space in the cache. We should set a cutoff for how many group by fields we still think make sense to cache.

If we wouldn't set cacheable=true, we would use the `sendQuery` function in `SiloClient.kt`, which streams responses directly instead of materializing the response.

fhennig:
resolves #1849

Aggregations that group by more than 20 columns (plain fields plus sequence-position fields) are no longer cached. The cache materializes the full result into a List to store it; for high-cardinality group-bys this can exhaust the heap (and did, causing a LAPIS outage). Such queries are now streamed straight through instead. Their cache hit rate is close to zero anyway since each filter combination is unique.

## PR Checklist
- ~~All necessary changes are explained in the `llms.txt`.~~

### DIFF
--- a/lapis/src/main/kotlin/org/genspectrum/lapis/silo/SiloQuery.kt
+++ b/lapis/src/main/kotlin/org/genspectrum/lapis/silo/SiloQuery.kt
@@ -37,6 +37,15 @@ interface CommonActionFields {
 
 const val ORDER_BY_RANDOM_FIELD_NAME = "random"
 
+/**
+ * Aggregations that group by more than this many columns (plain fields plus sequence-position fields)
+ * are not cached. Their result cardinality grows roughly with the number of matching sequences, so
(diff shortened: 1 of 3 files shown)


fengelniederhammer (approved):
Looks reasonable 👍

PR #1853, opened 2026-09-03
fix(lapis): expire siloQueryCache entries 30h after last access

fhennig:
The query cache was only ever cleared on a SILO data version change, which can be days apart. Add a 30h idle expiry so heap held by one-off queries (e.g. an analytics script sweeping many filter combinations) is released instead of lingering until the next data update.

This is a minimal, low-risk step. Bounding the cache by the memory size of its entries instead of by entry count is tracked separately.

## PR Checklist
- [ ] All necessary changes are explained in the `llms.txt`.

### DIFF
--- a/lapis/src/main/resources/application.properties
+++ b/lapis/src/main/resources/application.properties
@@ -4,7 +4,9 @@ springdoc.swagger-ui.operationsSorter=alpha
 server.forward-headers-strategy=framework
 
 spring.cache.cache-names=siloQueryCache
-spring.cache.caffeine.spec=maximumSize=50000,softValues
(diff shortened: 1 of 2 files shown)


fengelniederhammer (commented):
Why do we need to expire those queries? In my understanding those are evicted anyway when the cache is full and new entries come in.

fhennig:
Ok, we decided to not do this, as we're not sure if it actually helps or by how much, and we'd rather keep it simple for now. Ideally we'd use a weighted cache later on, so we can get rid of the `softValues`, that would be the real fix.

Issue #682, opened 2024-03-01
Info endpoint is slow


chaoran-chen:
The `/sample/info` on the production server is quite slow and needs >600 ms. Does anyone know why?

![image](https://github.com/GenSpectrum/LAPIS/assets/18666552/67636c9f-df8d-43d1-93d1-05e6f805d03a)

Request IDs:

- 1412b2a2-6e45-4702-93d0-fe7dafa99d71
- 4082e46e-ee4b-49f7-855d-618f5ad71f9c
- 82a550e1-802c-4221-8b83-4098e498b467
- c72c24f2-ae34-45aa-a174-91e047346889
- 5e481acd-3d00-4890-979e-0f90a63a7b37

PR #1866, opened 2026-09-10
feat(lapis): make the SILO info timeout configurable

corneliusroemer:
callInfo() has had a hard-coded 100 ms request timeout since #1593, picked so that the health endpoint cannot block when SILO does not respond. That same budget governs /sample/info, so a SILO that is merely a little slow (takes longer than 100ms) rather than down makes LAPIS return a 500. Users would not mind at all waiting say 200ms.

Rather than making a new judgment on what the timeout should be (there's a tradeoff between how long health can take if unhealthy and how slow silo can be before lapis returns 500) I propose to make the timeout configurable. So nothing changes unless users override.

Loculus would use this override to avoid flakes when SILO is just a bit slow (related to https://github.com/loculus-project/loculus/pull/7298)

There's a separate issue about why info is sometimes slow: https://github.com/GenSpectrum/LAPIS/issues/682 - if that's still the case then 100ms is definitely too short and the default should be raised.

Claude-Session: https://claude.ai/code/session_01PRvw1U8cSHWotT9kL5r8FC


corneliusroemer:
Superseded by #1868 - don't think it needs to be configurable, the timeout just needs to be restricted to where it's intended

PR #1867, opened 2026-09-10
feat(lapis): distinguish a SILO timeout from SILO being unreachable

corneliusroemer:
A timeout was reported as "Could not connect to silo", which is wrong and misleads whoever reads it. HttpTimeoutException means LAPIS did connect, sent the request and gave up waiting - SILO is reachable and may be answering other callers perfectly well. Only ConnectException means not reachable.

Related to #1866 in that the misleading exception came up while figuring out what caused info to return silo connection error.

SiloTimeoutException maps to 503, the same status the other SILO-unavailable paths use.

Also edits the message for an uncaught silo exception: the problem isn't necessarily connection failure, could be anything so message shouldn't assume.

## PR Checklist


corneliusroemer:
Superseded by #1871

PR #1868, opened 2026-09-11
fix(lapis): only timeout /sample/info calls in SILO health indicator's calls

corneliusroemer:
#1593 added a 100ms timeout for all calls to SILO's `/sample/info` because the silo health indicator which is used by by general health endpoint started using that function and health response shouldn't block indefinitely on the SILO call.

This fix had unintended side effects: all other users of that function now have the same very short timeout limitation of 100ms even if they don't mind waiting: they suddenly get a SILO unreachable exception even if SILO is healthy just a bit slow.

Loculus is affected by this (unintended) side effect: /sample/info starts randomly failing when there's high load.

This PR preserves the intent of #1593 while removing the unintended side effects: only the health endpoint sets a timeout.

Whether or not other callers to `/sample/info` should use a timeout and what that timeout should be can be decided separately.

### DIFF
--- a/lapis/src/main/kotlin/org/genspectrum/lapis/health/SiloHealthIndicator.kt
+++ b/lapis/src/main/kotlin/org/genspectrum/lapis/health/SiloHealthIndicator.kt
@@ -7,6 +7,9 @@ import org.springframework.boot.health.contributor.Health
 import org.springframework.boot.health.contributor.HealthIndicator
 import org.springframework.boot.health.contributor.Status
 import org.springframework.stereotype.Component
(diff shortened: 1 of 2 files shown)


fhennig (approved):
Thank you for this PR, I think it makes sense.

Besides the heath indicator, this function is used in `/sample/info` and in the `DataVersionCacheInvalidator.invalidateSiloCache()`, where we can also wait longer.

We mainly wanted the health indicator to respond quickly, even if SILO is unreachable.

This makes sense to me and thanks for the fix and describing the issue in the PR description!

PR #1869, opened 2026-09-11
chore(github-actions): pin the Node version CI runs on

corneliusroemer-agent:
`lts/*` at every `setup-node` step means the Node major CI runs on is whatever npm last reclassified as LTS. The upgrade (automatic by Github CI) then lands with no diff, in an unrelated run, and anything that depended on the old runtime breaks in a way that looks like a bug in this repo.

That caused waste of time on investigation in https://github.com/GenSpectrum/LAPIS/issues/1423 ("I spent 3 hours trying to debug this last week, but couldn't do it")

Solution is to pin this currently unpinned dep:

- One `.nvmrc` at the repo root, read by all three workflows via `node-version-file`. A Node upgrade becomes a one-line change someone proposes on purpose (note: dependabot won't do this)
- `engines: { node: ">=24" }` in `lapis-e2e` as the floor the tests actually need (not `engine-strict`, which would apply to the whole dependency tree) - possibly worth using as a pin on major version rather than just lower bounded.
- A mocha require hook that fails the e2e suite up front on an older Node, with a message naming the cause. That's the part that saves the debugging time: the suite's correctness is Node-coupled either way, so refusing to run beats letting a correct LAPIS produce confusing assertion failures.

No behaviour change today — `lts/*` already resolves to 24, so this only removes the silent-drift path. Note that nothing will auto-bump `.nvmrc`; Dependabot has no ecosystem for it, which is the point, but it does mean the next Node major needs a deliberate PR.


lapis-e2e/test/common.spec.ts
```diff
@@ -220,7 +220,7 @@ describe('All endpoints', () => {
         expect(response.headers.get('content-type')).to.equal(expectedContentType(route.servesType));
         expect(response.headers.get('content-encoding')).equals('zstd');
 
-        // fetch automatically decompresses zstd responses as of node 24
+        // fetch automatically decompresses zstd responses as of node 24 (the version pinned in .nvmrc)
```
corneliusroemer:
```suggestion
        // fetch automatically decompresses zstd responses as of node 24
```

fengelniederhammer (approved):
Thanks, LGTM 👍

PR #1871, opened 2026-09-11
feat(lapis): more informative SILO exception handling and error messages

corneliusroemer-agent:
> [!IMPORTANT]
> **Stacked on #1868 and #1870 — please merge those first.**
> This branch is based on #1870, so the diff here currently contains their commits too. 
>
> To see **only** this PR's change, diffed against the branch it is stacked on:
> https://github.com/corneliusroemer-agent/LAPIS/compare/silo-info-timeouts...silo-timeout-exception

## Why

Every failure to reach SILO except a refused connection was reported with the same sentence: "Could not connect to silo" when in fact the underlying problem might be something else. Different possible failures:
- A SILO request that ran out of time was reported as a connection failure when it could just be slow silo.
- `UnknownHostException`, `NoRouteToHostException` and `PortUnreachableException` fell into the catch-all, became a plain `RuntimeException`, and were answered with a **500** instead of the **503** that a refused connection gets
- The catch-all claimed a cause it does not know. By construction it is the branch for failures we could not classify, so "Could not connect to silo" is a guess presented as a fact.

## What changes for someone reading logs or health output

| situation | before | after |
| --- | --- | --- |
| request outlived its timeout | "Could not connect to silo", 503 | "Timed out waiting for a response from silo at `<uri>` after `<n>`ms", 503 |
| DNS / no route / port unreachable | "Could not connect to silo", **500** | "Could not connect to silo at `<uri>`", **503** |
| anything else | "Could not connect to silo", 500 | "Error talking to silo at `<uri>`", 500 |

`/actuator/health` changes too: A timeout used to fall through to the indicator's catch-all and report `"Unexpected error checking SILO"` with `siloStatus: DOWN`. But a call exceeding 100 ms is no certain evidence that SILO is down, just that it's slow on this particular call. It now reports `siloStatus: UNKNOWN`. The component itself stays `UP` as before — LAPIS's own health never depended on SILO — so no status code and no probe behaviour changes. A DNS or routing failure used to reach that same catch-all and now correctly says "SILO not reachable".

Two log lines were actively unhelpful during exactly the incidents you would want them for. `QueryParseModel` warned "Could not get current SILO data version" and discarded the cause, so it could not tell you whether SILO was slow, refused, or erroring; it now includes it, and after #1870 that is the warning most likely to appear. The data version poller logged its failures at `debug`, which means a SILO that merely answers slowly stops cache invalidation entirely with nothing visible at the default log level; it warns now.

## Judgement calls

`UNKNOWN` rather than `DOWN` for a timeout in the health details, on the grounds that 100 ms elapsing tells you nothing certain about SILO's state. `DOWN` is defensible if anything out there alerts on that field and you would rather it not go quiet.

## Not done here

Cache eviction is deliberately untouched. The poller substitutes a sentinel data version, and so clears the cache, only for `SiloUnavailableException`. `SiloTimeoutException` and `SiloNotReachableException` both extend `Exception` directly, so neither evicts — exactly as before this change. Evicting on a timeout in particular would be harmful: it would dump the cache precisely when SILO is already struggling, and send every subsequent query straight back at it.

https://claude.ai/code/session_0164mSDmZH2xfds5wZr3VVLx

### DIFF
--- a/lapis/src/main/kotlin/org/genspectrum/lapis/controller/ExceptionHandler.kt
+++ b/lapis/src/main/kotlin/org/genspectrum/lapis/controller/ExceptionHandler.kt
@@ -10,6 +10,7 @@ import org.genspectrum.lapis.response.LapisErrorResponse
 import org.genspectrum.lapis.response.LapisInfoFactory
 import org.genspectrum.lapis.silo.SiloException
 import org.genspectrum.lapis.silo.SiloNotReachableException
+import org.genspectrum.lapis.silo.SiloTimeoutException
 import org.genspectrum.lapis.silo.SiloUnavailableException
 import org.springframework.boot.autoconfigure.web.WebProperties
 import org.springframework.boot.webmvc.autoconfigure.error.BasicErrorController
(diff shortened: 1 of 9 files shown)


fengelniederhammer:
Done in #1882

